PR #1, opened 2023-01-11
Fix: Fix logging scenarios when no config was passed

SebastianMorawiec:
# The problem

After config reorganisation, user could no longer do orq login -s <uri> without specifying config.

# This PR's solution

Solution to the problem: When user uses orq login -s <uri> -c <config>, existing config is updated
When user uses orq login -s <uri> new config is created with auto-generated name

# Checklist


alexjuda (approved):
LGTM! ~It would be nice to have a test that verifies the behavior in absence of config file, but I suppose it's gonna be way easier to add that after CLI rework is done.~

edit: just found out the tests are there. Nevermind!

PR #4, opened 2023-01-11
exp: command for wf results

alexjuda:
# The problem

Getting workflow results is one of basic ways of seeing the effect of running workflows on Orquestra.

# This PR's solution

Adds new CLI command `dorq wf results [wf run ID] [-c config] [--download-dir]`:
* `dorq wf results` asks for config and wf run to show. Gets wf output values from the runtime. For each one, prints object type and a best-effort pretty print.
* `dorq wf results --download-dir data/` also asks for config and wf run. However, dumps the artifact values as files under `data/<wf run id>_<index>.<ext>`. Serialization format is selected using similar heuristics as on QE. The extension matches the format used. The created files only contain the serialized value, but no JSON envelope. If the artifact value is an `object()`, the file would be binary, with the contents being pickle bytes.

# Checklist


src/orquestra/sdk/_base/cli/_dorq/_dumpers.py
```diff
@@ -0,0 +1,89 @@
...
+"""
+Code that stores values on disk as a result of a CLI command.
+"""
+import json
+import typing as t
+from dataclasses import dataclass
+from functools import singledispatch
+from pathlib import Path
+
+import dill
```
jamesclark-Zapata, line 14: **issue:** ‏Using `dill` directly here would make more annoying to switch out the library later. Can you use the [`serde`](https://github.com/zapatacomputing/orquestra-workflow-sdk/blob/main/src/orquestra/sdk/_base/serde.py) module? Or add the missing functionality there?

src/orquestra/sdk/_base/cli/_dorq/_repos.py
```diff
@@ -101,6 +101,33 @@ def stop(self, wf_run_id: WorkflowRunId, config_name: ConfigName):
...
+            orquestra.sdk.exceptions.WorkflowRunNotSucceeded: when the workflow is no
+                longer executing, but wasn't succeded.
+        """
+        try:
+            wf_run = sdk.WorkflowRun.by_id(wf_run_id, config_name)
+        except (exceptions.NotFoundError, exceptions.ConfigNameNotFoundError):
+            raise
+
+        try:
+            outputs = wf_run.get_results(wait=True)
```
jamesclark-Zapata, line 125: **question:** ‏Do we always want this to block? This may look like a broken CLI if we don't show anything

src/orquestra/sdk/_base/cli/_dorq/_entry.py
```diff
@@ -93,6 +98,43 @@ def view(wf_run_id: t.Optional[str], config: t.Optional[str]):
...
+def results(
+    wf_run_id: t.Optional[str],
+    config: t.Optional[str],
+    download_dir: t.Optional[Path],
+):
+    """
+    Shows preview of a workflow output values corresponding to the variables
+    returned from the ``@workflow`` function. This tries to print a human-friendly
+    values preview, but the output isn't guaranteed to be a valid parseable value. If
+    you need the artifact value for further processing, use the ``download_dir`` option
```
jamesclark-Zapata, line 124: **nitpick:** ‏We use `artifact` as the outputs of each task, atm.

wdyt about:
```suggestion
    you need the result value for further processing, use the ``download_dir`` option
```

src/orquestra/sdk/_base/_api.py
```diff
@@ -494,6 +495,8 @@ def get_results(self, wait: bool = False) -> t.Sequence[t.Any]:
             WorkflowRunNotStarted: when the workflow run has not started
             WorkflowRunNotFinished: when the workflow run has not finished and `wait` is
                                    False
+            WorkflowRunNotSucceeded: when the workflow is no longer executing, but it wasn't
+                succeded.
```
jamesclark-Zapata, line 499:
```suggestion
            WorkflowRunNotSucceeded: when the workflow is no longer executing, but it did not
                succeed.
``

File: src/orquestra/sdk/_base/cli/_dorq/_dumpers.py [hunk 5]
```diff
@@ -0,0 +1,89 @@
...
+
+        Creates missing directories. Generates filenames based on ``wf_run_id`` and
+        ``output_index``. Figures out the serialization format based on the object. The
+        generated file extension matches the inferred format.
+
+        No standard errors are expected to be raised.
+        """
+        # Dumping to strings instead of directly to byte buffer is suboptimal, but we
+        # wanna use the same serialization procedure as Corq and QE.
+        result_obj = serde.result_from_artifact(
```
jamesclark-Zapata, line 79: **question:** ‏‏We do deserialisation (python API) -> serialisation (here) -> deserialisation (`_dump_result`) -> serialisation (`_dump_result`). Is there a better way to handle this?

jamesclark-Zapata (changes requested):

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 👩‍🔬 *Experimental*
 
+* New CLI command: `python -m orquestra.sdk._base.cli._dorq._entry workflow results`. It shows a preview of the workflow run output artifact. Can download the artifact values to the provided dir.
```
jamesclark-Zapata, line 13:
```suggestion
* New CLI command: `python -m orquestra.sdk._base.cli._dorq._entry workflow results`. It shows a preview of the workflow run output artifact and can download the result values to the provided directory.
* ```

src/orquestra/sdk/_base/cli/_dorq/_repos.py
jamesclark-Zapata: Raising an error (which `get_results` does) should be enough?

src/orquestra/sdk/_base/cli/_dorq/_dumpers.py
jamesclark-Zapata: Can you right down the functions calls of how you might do this?

alexjuda: I tried many different approaches to reuse as much serde as possible and ended up with f906729125126cef7604f999e0ec34b4.

It doesn't use the same code path as QE but at least we have `dill.dumps()` and `dill.dump()` in the same module.

src/orquestra/sdk/_base/cli/_dorq/_repos.py
alexjuda: Changed in 19d53fc

src/orquestra/sdk/_base/cli/_dorq/_dumpers.py
jamesclark-Zapata: I think it's OK for now but my issue is we're re-serialising still, so if we ever allow a user to pick the serialisation we may overwrite their choice.

Can you add a ticket to tackle this later? I'm thinking that maybe we could add a `WorkflowRun.get_results(raw=True)` which skips deserialisation.

jamesclark-Zapata (approved):

PR #10, opened 2023-01-13
fix: orq up timing out

alexjuda:
# The problem

Sometimes `dorq up` raised timeout errors on my machine. Even when if it succeeds, the command time is dangerously close to our timeout threshold. I'm using Python 3.10.2.

```
$ time dorq up
  [####################################]  100%
Ray  Running  Started!

real    0m11.279s
user    0m2.972s
sys     0m0.894s
```

# This PR's solution

Increases the timeout threshold for `dorq {up,down,services}` by the factor of 2. Let our future selves deal with the root problem.

# Checklist

- [ ] The [changelog file](CHANGELOG.md) has been updated with a user-readable description of the changes (if the change isn't visible to the user in any way, check the box).

### DIFF
--- a/src/orquestra/sdk/_base/_services.py
+++ b/src/orquestra/sdk/_base/_services.py
@@ -88,7 +88,7 @@ def _compose_path():
 
 
 # Timeout for inter-process commands (seconds).
-IPC_TIMEOUT = 10
+IPC_TIMEOUT = 20
 
 
 class DockerException(Exception):


SebastianMorawiec (approved):
Approved, but can you also check how long does the command ray start --head --storage=/tmp/ray takes? just so we know its mostly on ray not on us for that timeout

alexjuda:
@SebastianMorawiec 

- `time ray up ...`: around 7.161s
- `time dorq up`: between 5.7s and 8s, but it's not very stable. I can easily imagine it taking ~11s, eg. after a reboot.
- `time dorq up` when I replaced the the `ray up` with `ls` for the subprocess: around 3.448s

SebastianMorawiec:
> 

Definitely - if ray up takes 7 sec, 10 sec timeout is too low as there easily might be some hick-up.

PR #13, opened 2023-01-16
Feat: Prompt users for configs in all DORQ commands

SebastianMorawiec:
# The problem

New DORQ commands are prompt-based when the user doesnt pass the value. Configs had to be passed explicitly for now in orq submit command

# This PR's solution

Prompt user for configs in dorq submit
Also remove duplicate configs from listings (ray and local - they were only aliases on each other)
Remove warning of naming in process config when submitting wf in-process

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/cli/_dorq/_workflow/_submit.py
+++ b/src/orquestra/sdk/_base/cli/_dorq/_workflow/_submit.py
@@ -8,7 +8,7 @@ import typing as t
 
 from orquestra.sdk import exceptions
 
-from .. import _repos
+from .. import _arg_resolvers, _repos
 from .._ui import _presenters, _prompts
 
 
@@ -26,6 +26,7 @@ class Action:
         presenter=_presenters.WrappedCorqOutputPresenter(),
         wf_def_repo=_repos.WorkflowDefRepo(),
         wf_run_repo=_repos.WorkflowRunRepo(),
+        config_resolver: t.Optional[_arg_resolvers.ConfigResolver] = None,
     ):
         # text IO
         self._prompter = prompter
@@ -34,6 +35,9 @@ class Action:
         # data sources
         self._wf_run_repo = wf_run_repo
         self._wf_def_repo = wf_def_repo
+        self._config_resolver = config_resolver or _arg_resolvers.ConfigResolver(
+            wf_run_repo=wf_run_repo, prompter=prompter
+        )
 
     def on_cmd_call(
         self, module: str, name: t.Optional[str], config: t.Optional[str], force: bool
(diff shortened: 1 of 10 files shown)


src/orquestra/sdk/_base/cli/_dorq/_workflow/_submit.py
```diff
@@ -54,13 +58,7 @@ def _on_cmd_call_with_exceptions(
...
         """
         # 1. Resolve config
-        # TODO: prompt user for selecting config.
-        # Jira ticket: https://zapatacomputing.atlassian.net/browse/ORQSDK-674
-        if config is None:
-            raise ValueError(
-                "Temporarily, 'config' needs to be passed explicitly. Sorry!"
-            )
-        resolved_config = config
+        resolved_config = self._config_resolver.resolve(None, config)
```
alexjuda, line 61: **non-blocking**: the previously implemented config resolver was intended for use with commands that require both `wf-run-id` and `config`. Submit doesn't use `wf-run-id` (this value is generated as a result of submission). Using `self._config_resolver.resolve(None, config)` _accidentally_ works because of the fact that we pass through the `config` value even if we don't pass wf run ID to `orq wf view`, which an implementation detail of the resolver for other commands and might easily break. 

WDYT about making a separate arg resolver just for `orq wf submit`? Internally it can wrap the current `ConfigResolver` but we would have a clear place where we make the assumption about internal workings of the resolution.

tests/cli/dorq/test_repos.py
```diff
@@ -260,21 +261,82 @@ def test_list_wf_run_ids(monkeypatch):
 
 
 class TestConfigRepo:
+    class TestUnitConfigRepo:
```
alexjuda, line 264: The enclosing class already provides a namespace with `ConfigRepo` in it.
```suggestion
    class TestUnit:
```

tests/cli/dorq/test_repos.py
```diff
@@ -260,21 +261,82 @@ def test_list_wf_run_ids(monkeypatch):
...
+            config_parameter = ""
+            runtime_parameter = ""
+            options_parameter = {}
+
+            # Check parameters passed to _Config
+            def validate_save_parameters(config, runtime, options):
+                nonlocal config_parameter, runtime_parameter, options_parameter
+                config_parameter = config
+                runtime_parameter = runtime
+                options_parameter = options
```
alexjuda, line 304: **suggestion**: did you consider using mocks as spies?

alexjuda (approved):
Added non-blocking suggestions :)

tests/cli/dorq/test_repos.py
SebastianMorawiec: I did, but as this is not dependency-injection object its kinda herder to do, if you feel like a better idea I can go with it, but this way its way easier and simpler + it actually verifies it being called and checks all the stuff better than Mocks as Ive currently found out

alexjuda: Hooking up a mock here would be similar to hooking up a function.

```
mock_save_or_update = Mock()
monkeypatch.setattr(
    sdk._base._config, "save_or_update", mock_save_or_update
)
```

Then, you can check arguments that the mock was called with:
```
(config_parameter,
 runtime_parameter,
 options_parameter) = mock_save_or_update.call_args[0]
```

PR #15, opened 2023-01-16
fix: Disable Ray logging via runtime config

jamesclark-Zapata:
# The problem

Ray can sometimes print out a log of unrelated messaged when we use it.
This is confusing for users in Jupyter notebooks because these messages end up in red boxes because it is printed to std err.

# This PR's solution

Switches Ray's logger to log level ERROR and disables Ray's configuration of its logger, with the option to leave it configured.

# Checklist

### DIFF
--- a/src/orquestra/sdk/_ray/_dag.py
+++ b/src/orquestra/sdk/_ray/_dag.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import dataclasses
 import json
+import logging
 import re
 import traceback
 import typing as t
@@ -562,6 +563,7 @@ class RayParams:
     log_to_driver: bool = True
     storage: t.Optional[t.Union[str, _client.Storage]] = None
     _temp_dir: t.Optional[str] = None
+    configure_logging: bool = True
 
 
 # Defensive timeout for Ray operations that are expected to be instant.
@@ -599,6 +601,7 @@ class RayRuntime(RuntimeInterface):
             log_to_driver=config.runtime_options["log_to_driver"],
             storage=config.runtime_options["storage"],
             _temp_dir=config.runtime_options["temp_dir"],
+            configure_logging=config.runtime_options["configure_logging"],
         )
         cls.startup(ray_params)
         return RayRuntime(client=client, config=config, project_dir=project_dir)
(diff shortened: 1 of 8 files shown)


src/orquestra/sdk/_ray/_dag.py
```diff
@@ -562,6 +563,7 @@ class RayParams:
     log_to_driver: bool = True
     storage: t.Optional[t.Union[str, _client.Storage]] = None
     _temp_dir: t.Optional[str] = None
+    configure_logging: bool = True
```
alexjuda, line 566: **thought**: you might need to also extend some test fixtures. I can't remember now (sorry, it's late!) but if the CI failure logs will contain hints.

alexjuda (approved):
Nice!

SebastianMorawiec:
Sorry for jumping in late, but I'd like to add that after recent config changes Im not a huge fan of extending configs with anything. I'd rather either disable it entirely or leave it up for dorq up ray command (or idk... env variable?). Especially that we no longer hold Ray config on disk, so there is no way to actually modify this value.

jamesclark-Zapata:
> Sorry for jumping in late, but I'd like to add that after recent config changes Im not a huge fan of extending configs with anything. I'd rather either disable it entirely or leave it up for dorq up ray command (or idk... env variable?). Especially that we no longer hold Ray config on disk, so there is no way to actually modify this value.

You could modify this value by making a new config and putting it in the config file.

What's your reasoning for not wanting to add a new file?

SebastianMorawiec:
> > Sorry for jumping in late, but I'd like to add that after recent config changes Im not a huge fan of extending configs with anything. I'd rather either disable it entirely or leave it up for dorq up ray command (or idk... env variable?). Especially that we no longer hold Ray config on disk, so there is no way to actually modify this value.
> 
> You could modify this value by making a new config and putting it in the config file.
> 
> What's your reasoning for not wanting to add a new option?

But you cannot create new RAY RUNTIME config anymore. I do not like expanding settings as those settings gives absolute no purpose whatsoever. They might be modified by advanced users like 1 in a 1000 usages, and keeping the configs as they are right now is a big pitfall.

BUT as this would be just another unused option for config, its not like it complicates things, If we ever remove those stuff entirely this wont make it more difficult so its not like im strictly opposing to merge it

PR #16, opened 2023-01-16
internal: machinery for task run identification

alexjuda:
# The problem

A couple of `dorq` commands require identifying the task run. There's a couple of ways we can figure it out based on user input. The same code is likely to be used by different commands.

# This PR's solution

Implements Argument Resolver for finding task run ID.

# Checklist

- [x] The [changelog file](CHANGELOG.md) has been updated with a user-readable description of the changes (if the change isn't visible to the user in any way, check the box). I'm not adding a changelog entry because this PR's content will be only visible to the user when we implement task-related commands in the future.


alexjuda:
cc @SebastianMorawiec

src/orquestra/sdk/_base/cli/_dorq/_arg_resolvers.py
```diff
@@ -103,6 +104,108 @@ def resolve(
...
+            else:
+                resolved_fn_name = fn_names[0]
+
+        inv_ids = self._wf_run_repo.get_task_inv_ids(
+            wf_run_id=wf_run_id,
+            config_name=config,
+            task_fn_name=resolved_fn_name,
+        )
+        if len(inv_ids) > 1:
+            resolved_inv_id = self._task_inv_prompter.choice(
```
SebastianMorawiec, line 152: Thought: (Non-blocking but something to be aware) - it came out on our latest OWG that we might want to be prompt-less for some automated-usage. Given this logic, when user provides fn_name there is no way to combine other things to avoid prompt. This makes sense given the nature of this parameter, but I think its worth saying it loud.

src/orquestra/sdk/_base/cli/_dorq/_repos.py
```diff
@@ -68,6 +78,22 @@ def get_wf_by_run_id(
...
+    def get_task_run_id(
+        self,
+        wf_run_id: WorkflowRunId,
+        task_inv_id: TaskInvocationId,
+        config_name: ConfigName,
+    ) -> TaskRunId:
+        wf_run_model = self.get_wf_by_run_id(
+            wf_run_id=wf_run_id, config_name=config_name
+        )
+        # TODO: figure out what happens when there's no matching task run.
```
SebastianMorawiec, line 90: Issue: hanging TODO. Either fix or create new Jira ticket and attach it here

src/orquestra/sdk/_base/cli/_dorq/_repos.py
```diff
@@ -20,7 +20,17 @@
...
+from orquestra.sdk.schema.ir import (
+    FunctionRef,
+    TaskDefId,
+    TaskInvocationId,
+    WorkflowDef,
+)
+from orquestra.sdk.schema.workflow_run import TaskRunId, WorkflowRun, WorkflowRunId
+
+
+def _find_first(f: t.Callable[[t.Any], bool], it: t.Iterable):
```
SebastianMorawiec, line 32: Thought: This is trivial function used once. Is it really needed? The signature confused me for a bit tbh 😂

src/orquestra/sdk/_base/cli/_dorq/_repos.py
```diff
@@ -133,6 +159,88 @@ def get_wf_outputs(self, wf_run_id: WorkflowRunId, config_name: ConfigName):
...
+        """
+        try:
+            wf_run = sdk.WorkflowRun.by_id(wf_run_id, config_name)
+        except (exceptions.NotFoundError, exceptions.ConfigNameNotFoundError):
+            raise
+
+        status_model = wf_run.get_status_model()
+        wf_def = status_model.workflow_def
+
+        # TODO: check when this is optional and add a message
```
SebastianMorawiec, line 180: Issue: hanging TODO. Either fix or create new Jira ticket and attach it here

src/orquestra/sdk/_base/cli/_dorq/_repos.py
```diff
@@ -133,6 +159,88 @@ def get_wf_outputs(self, wf_run_id: WorkflowRunId, config_name: ConfigName):
...
+        try:
+            wf_def = self._get_wf_def_model(wf_run_id, config_name)
+        except (exceptions.NotFoundError, exceptions.ConfigNameNotFoundError):
+            raise
+
+        # What we're really interested in are task_runs, but it's easier to test when
+        # we iterate over wf_defs's task_defs. Assumption: every task_def is being used
+        # in the workflow and corresponds to a task_run.
+        names_set = {
+            task_def.fn_ref.function_name for task_def in wf_def.tasks.values()
```
SebastianMorawiec, line 211: Nonblocking: thought: Is the function_name enough of an identifier? What if we have 2 functions in 2 different files/modules with exact same name. How would user distinguish between them? Or even 2 methods in 2 different classes

SebastianMorawiec: Btw. This is nonblocking for this PR, but still requires analysis. If you consider it out of scope, lets make Jira ticket for that

SebastianMorawiec (changes requested):
Well done, just remove hanging TODOs and would love to hear your comment of few minor thoughts.

src/orquestra/sdk/_base/cli/_dorq/_arg_resolvers.py
alexjuda: I think we have all the bases covered:
- `dorq task logs --fn-name foo` -> we prompt for config, wf run ID and task inv ID.
- `dorq task logs --fn-name foo --task-inv-id inv1` -> we prompt for config, and wf run ID.
- `dorq task logs --fn-name foo --task-inv-id inv1` -> we prompt for config, and wf run ID.
- `dorq task logs --wf-run-id wf1 --fn-name foo --task-inv-id inv1` -> we prompt for config (if we can't infer it from wfs DB cache).
- `dorq task logs --config ray --wf-run-id wf1 --fn-name foo --task-inv-id inv1` -> we don't prompt for anything
- `dorq task logs --config ray --task-run-id foobar` -> we don't prompt for anything

src/orquestra/sdk/_base/cli/_dorq/_repos.py
alexjuda: It communicates the intent better than the `next(filter(...))`. I believe it's worth it for the same reason why `map()` and comprehensions are better than the generic `for` loop.

src/orquestra/sdk/_base/cli/_dorq/_repos.py
alexjuda: It's a known limitation described in the docstring for `get_task_fn_names`:
```
        Extracts task function names used in this workflow run.

        If two different task defs have the same function name this returns a single
        name entry. This can happen when similar tasks are defined in two different
        modules.
```

I made an attempt to use a different way to identify the task defs to avoid shadowing (see af7d9c952a306cd313003fdc330cff7749899b4c and the revert coming right after it), and it's possible, but I couldn't figure out a user-friendly way of providing this as an CLI arg. Asking the user to provide a `task_def_id` would suck, defeating the whole purpose of these filters

alexjuda: > Is the function_name enough of an identifier?

There are situations where it's not, and that's why there's another inquiry step for disambiguating between task invocations :)

src/orquestra/sdk/_base/cli/_dorq/_repos.py
alexjuda: Other languages come with a standard library that includes a plethora of utilities for doing common things with sequences, like `find_first`, `window`, `batch`, etc. Python is kinda backwards in this regard 😄

src/orquestra/sdk/_base/cli/_dorq/_repos.py
alexjuda: Updated!

src/orquestra/sdk/_base/cli/_dorq/_repos.py
alexjuda: Fixed!

SebastianMorawiec (approved):
🚢 it

PR #19, opened 2023-01-17
fix: stop tests writing to workflows.db

BenjaminMummery:
# The problem

Running integration tests writes ~40 new entries to workflows.db

# This PR's solution

Mocks the .orquestra location for the integration tests.

# Checklist

### DIFF
--- a/tests/runtime/ray/conftest.py
+++ b/tests/runtime/ray/conftest.py
@@ -1,0 +1,23 @@
+################################################################################
+# © Copyright 2022 Zapata Computing Inc.
+################################################################################
+import os
+
+import pytest
+
+from orquestra.sdk._base._testing import _connections
+
+
+@pytest.fixture(scope="module")
+def change_test_dir(tmp_path_factory, request):
+    project_dir = tmp_path_factory.mktemp("project")
+    os.chdir(project_dir)
+    yield project_dir
+    os.chdir(request.config.invocation_dir)
+
+
+@pytest.fixture(scope="module")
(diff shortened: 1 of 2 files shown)


tests/runtime/ray/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+def shared_ray_conn():
+    with _connections.make_ray_conn() as ray_params:
+        yield ray_params
+
+
+@pytest.fixture(scope="module")
+def change_db_location(change_test_dir):
+    os.environ["ORQ_DB_LOCATION"] = os.path.join(change_test_dir, "db.db")
+    yield
+    del os.environ["ORQ_DB_LOCATION"]
```
jamesclark-Zapata, line 29: **suggestion (non-blocking):** ‏Wdyt of:
```suggestion
@pytest.fixture(scope="module")
def change_db_location(change_test_dir, monkeypatch):
    monkeypatch.setenv("ORQ_DB_LOCATION", os.path.join(change_test_dir, "db.db")
```

jamesclark-Zapata (approved):

tests/runtime/ray/conftest.py
SebastianMorawiec: @jamesclark-Zapata does it actually work? I thought monkeypatch is function-scoped fixture and wont work when used in module-scoped fixture

jamesclark-Zapata: I think you're right!

PR #20, opened 2023-01-18
chore: update arg resolvers and prompters

jamesclark-Zapata:
# The problem

In some cases, we need to list workflow runs in the CLI and then use the workflow run.

Previously, we listed the IDs (by getting the runs and then printing the IDs) and then calling the runtime to get the actual workflow run.

If a user cancelled a request in an arg resolver, we would get an ugly error message.

# This PR's solution

 - Adds a new resolver to either resolve an ID or a run
 - Adds an exception for when a user cancels the prompting in the arg resolvers.

# Checklist


alexjuda (approved):
LGTM!

PR #21, opened 2023-01-18
chore: Get workflow def in workflow run methods in CE driver

jamesclark-Zapata:
# The problem

We need the workflow defintion to drive some UI

# This PR's solution

- Updates the CE driver to grab the workflow def for a workflow run

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_driver/_models.py
+++ b/src/orquestra/sdk/_base/_driver/_models.py
@@ -168,7 +168,7 @@ class TaskRunResponse(pydantic.BaseModel):
         )
 
 
-class WorkflowRunResponse(pydantic.BaseModel):
+class MinimalWorkflowRunResponse(pydantic.BaseModel):
     """
     Implements:
         https://github.com/zapatacomputing/workflow-driver/blob/main/openapi/src/schemas/WorkflowRun.yaml#L1
@@ -176,15 +176,32 @@ class WorkflowRunResponse(pydantic.BaseModel):
 
     id: WorkflowRunID
     status: RunStatusResponse
+    definitionId: WorkflowDefID
+
+    def to_ir(self, workflow_def: Optional[WorkflowDef] = None) -> WorkflowRun:
+        return WorkflowRun(
+            id=self.id,
+            status=self.status.to_ir(),
+            task_runs=[],
+            workflow_def=workflow_def,
+        )
+
+
+class WorkflowRunResponse(MinimalWorkflowRunResponse):
+    """
+    Implements:
+        https://github.com/zapatacomputing/workflow-driver/blob/main/openapi/src/schemas/WorkflowRun.yaml#L1
+    """
+
     owner: str
     taskRuns: List[TaskRunResponse]
 
-    def to_ir(self) -> WorkflowRun:
+    def to_ir(self, workflow_def: Optional[WorkflowDef] = None) -> WorkflowRun:
         return WorkflowRun(
             id=self.id,
             status=self.status.to_ir(),
             task_runs=[t.to_ir() for t in self.taskRuns],
-            workflow_def=None,
+            workflow_def=workflow_def,
         )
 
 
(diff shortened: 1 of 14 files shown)


src/orquestra/sdk/schema/workflow_run.py
```diff
@@ -43,7 +43,7 @@ class TaskRun(BaseModel):
 
 class WorkflowRun(BaseModel):
     id: WorkflowRunId
-    workflow_def: t.Optional[WorkflowDef]
+    workflow_def: WorkflowDef
```
alexjuda, line 46: **praise**: yasss 🙌

src/orquestra/sdk/_base/_driver/_models.py
```diff
@@ -168,23 +168,40 @@ def to_ir(self) -> TaskRun:
...
     status: RunStatusResponse
+    definitionId: WorkflowDefID
+
+    def to_ir(self, workflow_def: WorkflowDef) -> WorkflowRun:
+        return WorkflowRun(
+            id=self.id,
+            status=self.status.to_ir(),
+            task_runs=[],
+            workflow_def=workflow_def,
+        )
```
alexjuda, line 187: **issue**: later on, in a different part of the codebase we might see weird behavior when we have a variable of type `WorkflowRun` with always-empty list of task runs inside of it. In the past, we used [WorkflowRunMinimal](https://github.com/zapatacomputing/orquestra-workflow-sdk/blob/bc81bde8270ffe533c893d2ebca8569c26b7c43b/src/orquestra/sdk/schema/workflow_run.py#L51) for situations where we don't have some of the details. Did you consider returning the `*Minimal` object here?

I know `WorkflowRunMinimal` doesn't have a `workflow_def` field, but is that a problem? Do we ever need wf def from `WorkflowRunMinimal`?

tests/sdk/v2/driver/resp_mocks.py
```diff
@@ -125,39 +127,53 @@ def make_submit_wf_run_response(id_: str):
...
+        "data": [
+            _wf_run_resp(id_, wf_def_id, status, [])
+            for id_, wf_def_id, status in zip(ids, workflow_def_ids, statuses)
+        ]
     }
 
 
-def make_list_wf_run_paginated_response(ids: List[str], statuses: List[RunStatus]):
+def make_list_wf_run_paginated_response(
+    ids: List[str], workflow_def_ids: List[str], statuses: List[RunStatus]
```
alexjuda, line 165: **suggestion,non-blocking**: there's a couple of different IDs here, with different semantics. Wanna add typealiases for them, in a similar way to `TaskRunID` vs `TaskInvocationID`?

alexjuda (changes requested):

src/orquestra/sdk/_base/_driver/_models.py
alexjuda: >  If we want to build the public API version of WorkflowRun, we need a workflow definition.

Could post an example of the public API where the `MinimalWorkflowRun` with wf def inside would be useful?

alexjuda: Thanks. 
* `_api.WorkflowRun.__init__` and `_api.TaskRun.__init__` indeed only need the wf_run_id and wf_def, but not other fields from `WorkflowRun` model like the wf status, and all the task run details. 
* If we don't wanna populate all the `workflow_run.WorkflowRun`'s in the objects returned from `sdk.list_workflow_runs()` we shouldn't mark the output type as list of `workflow_run.WorkflowRun`. It's like marking a variable type as a non-optional `str` and setting it to `""` or `"empty"`.

Sounds like we need something `workflow_run.WorkflowRunMinimal` with a `wf_def` field. How about third type, `workflow_run.WorkflowRunSummary`?

alexjuda: Yeah and yeah. I love the idea with a protocol, this will make it more open ended for the future.

alexjuda: The link points to `.iter_logs()` instead of `.list_workflow_runs()` but I'm assuming it was a misclick

jamesclark-Zapata: @alexjuda Protocol would not play nice with the type checker and instead of spending a bunch of time and trying to get it to work I just used inheritance.

alexjuda (approved):

PR #23, opened 2023-01-18
fix!: inconsistent data shapes when getting intermediate artifacts

alexjuda:
# The problem

* The contents of the dict returned from `wf_run.get_results()` depended on the runtime type used. For Ray and QE, this was a map if [task inv id]:[artifact value]. For In-Process runtime this was [artifact ID]:[artifact value].
* `InProcessRuntime` didn't conform to `RuntimeInterface`. The types and semantics didn't always match.
* The docstrings for `sdk.WorkflowRun.get_results()` mentioned `TaskRunID`s but in practice this was never used. Instead, we would use `TaskInvocationID` or `ArtifactID` as the filter and in the returned dict.

# This PR's solution

* Marks `InProcessRuntime` as an implementation of `RuntimeInterface`.
* Fixes invalid data shape returned from `InProcessRuntime.get_available_outputs()`.
* Fixes docstrings for `sdk.WorkflowRun.get_results()`.
* Simplifies `sdk.WorkflowRun.get_results()`: now it doesn't accept any arguments. Turns out it would be increasingly difficult to test after this PR + the client-side filtering seems to have limited user value.
* Fixes type annotations & updates tests for `CERuntime.get_available_outputs()`

# Checklist


src/orquestra/sdk/_base/_in_process_runtime.py
```diff
@@ -125,12 +143,31 @@ def get_workflow_run_outputs_non_blocking(
     ) -> t.Sequence[t.Any]:
         return self.get_workflow_run_outputs(workflow_run_id)
 
-    def get_available_outputs(self, workflow_run_id: WfRunId) -> t.Dict[str, t.Any]:
-        workflow_run_outputs = self.get_workflow_run_outputs(workflow_run_id)
-        output_task_ids = self._workflow_def_store[workflow_run_id].output_ids
-        return dict(zip(output_task_ids, workflow_run_outputs))
+    def get_available_outputs(
```
SebastianMorawiec, line 146: Issue: The rest of this PR is mostly docstring and interface changes already tested. But this function is completely rewritten. Could you provide at least some basic tests that it works as expected?

src/orquestra/sdk/_base/_in_process_runtime.py
```diff
@@ -125,12 +143,31 @@ def get_workflow_run_outputs_non_blocking(
...
 
-    def get_full_logs(self, _) -> t.Dict[str, t.List[str]]:
+            inv_outputs[inv.id] = output_vals
+
+        return inv_outputs
+
+    def get_full_logs(
+        self, run_id: t.Union[WorkflowRunId, TaskRunId, None] = None
+    ) -> t.Dict[str, t.List[str]]:
         return {}
```
SebastianMorawiec, line 171: suggestion: maybe throw an exception here. Its better than silently return always nothing

src/orquestra/sdk/_base/_in_process_runtime.py
```diff
@@ -224,3 +261,13 @@ def list_workflow_runs(
                 -limit:
             ]
         return wf_runs
+
+    @classmethod
+    def from_runtime_configuration(cls, *args, **kwargs):
+        raise NotImplementedError()
+
+    def get_all_workflow_runs_status(self, *args, **kwargs):
```
SebastianMorawiec, line 269: suggestion: maybe better exception? Meaning like "method {} is not implemented for in-process runtime". WDYT

SebastianMorawiec (changes requested):

src/orquestra/sdk/_base/_in_process_runtime.py
alexjuda: Yeah sure. TBH I submitted the PR just to see which tests are failing and I'm amazed the whole suite passed.

tests/sdk/v2/driver/test_ce_runtime.py
```diff
@@ -436,38 +436,12 @@ def test_happy_path(
...
         )
         mocked_client.get_workflow_run_artifact.assert_has_calls(
-            [call("1"), call("2"), call("3")]
+            [call("wf-art-1"), call("wf-art-2"), call("wf-art-3")]
         )
-        assert len(results) == 2
-        assert len(results["task-1"]) == 2
-        assert len(results["task-2"]) == 1
-
-    def test_map_remote_to_local_ids(
```
alexjuda, line 445: This test seemed to duplicate the prev one, so I merged them.

src/orquestra/sdk/_base/_api.py
```diff
@@ -557,11 +560,12 @@ def get_artifacts(
                             cannot be found, either because it does not exist or has not
                             been completed yet.
         Returns:
-            A dictionary of the task run artifacts with the shape::
-                task_run_id: values returned from the task
+            A dictionary with an entry for each task run in the workflow. The key is the
+                task's invocation ID. The value is whatever the task returned. If the
+                task has 1 output, it's the dict entry's value. If the tasks has n
```
alexjuda, line 565: [note to self]: make sure this is indeed the case

src/orquestra/sdk/_base/_in_process_runtime.py
alexjuda: Added :)

src/orquestra/sdk/_base/_in_process_runtime.py
alexjuda: Added!

src/orquestra/sdk/_base/_api.py
```diff
@@ -396,15 +396,16 @@ def start(self):
...
         """Block until the workflow run finishes.
 
         This method draws no distinctions between whether the workflow run completes
         successfully, fails, or is terminated for any other reason.
 
         Args:
-            frequency (float, optional): The frequence in Hz at which the status should
-                be checked. Defaults to 1.
+            frequency: The frequence in Hz at which the status should be checked.
+                Defaults to 1.
```
SebastianMorawiec, line 407: Question: isnt this default comment a lie? frequency is 0.25 default

src/orquestra/sdk/_base/_api.py
```diff
@@ -429,6 +430,8 @@ def wait_until_finished(self, frequency: float = 0.25) -> State:
         while status == State.RUNNING or status == State.WAITING:
             time.sleep(1.0 / frequency)
             status = self.get_status()
+            if print_status:
```
SebastianMorawiec, line 433: Thought: This is kinda missleading. It only prints status when its not ready (failed or succeeded).

SebastianMorawiec (approved):
I approve with 2 non-blocking things, although Im not a fan of this change for get_artifacts function tbh

src/orquestra/sdk/_base/_api.py
alexjuda: Ohhh nice catch! That's actually my prototyping code I forgot to roll back.

PR #25, opened 2023-01-23
Feat: Bump ray to 2.2

SebastianMorawiec:
# The problem

We should try to keep up with the latest version of ray. Also newer version of ray introduced fix for the bug that we had to workaround in SDK

# This PR's solution

Bumps ray to latest 2* version

# Checklist

### DIFF
--- a/src/orquestra/sdk/_ray/_dag.py
+++ b/src/orquestra/sdk/_ray/_dag.py
@@ -158,6 +158,7 @@ def _make_ray_dag_node(
     ray_kwargs: t.Mapping[str, t.Any],
     pos_unpack_specs: t.Sequence[PosArgUnpackSpec],
     kw_unpack_specs: t.Sequence[KwArgUnpackSpec],
+    project_dir,
     user_fn_ref: t.Optional[ir.FunctionRef] = None,
 ) -> _client.FunctionNode:
     """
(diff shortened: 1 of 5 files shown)


alexjuda:
@jamesclark-Zapata 
> Also newer version of ray introduced fix for the bug that we had to workaround in SDK

What was the bug?

tests/runtime/performance/test_cli_perf.py
```diff
@@ -78,21 +79,32 @@ def orq_workflow_run(ray_cluster, orq_project_dir):
...
+                    "workflow-run-results",
+                    workflow_id,
+                    "-d",
+                    orq_project_dir,
+                    "-c",
+                    "local",
+                ]
+            )
+            break
+        except Exception:
```
alexjuda, line 104: I suppose the errors raised here will be related to the subprocesses. Can you make the except close more specific?

src/orquestra/sdk/_ray/_dag.py
```diff
@@ -158,6 +158,7 @@ def _make_ray_dag_node(
     ray_kwargs: t.Mapping[str, t.Any],
     pos_unpack_specs: t.Sequence[PosArgUnpackSpec],
     kw_unpack_specs: t.Sequence[KwArgUnpackSpec],
+    project_dir,
```
alexjuda, line 161: Our code in this file is already difficult to follow. Please add the type annotation and describe what it is for in the docstring
```suggestion
    project_dir: t.Optional[Path],
```

alexjuda (changes requested):
Thanks for figuring it out! Added few minor suggestions

alexjuda (approved):

PR #28, opened 2023-01-25
feat!: return all logs from WorkflowRun

alexjuda:
# The problem

* Getting logs for the whole workflow is a common scenario during development.
* Current API required the users to first grab `TaskInvocationId`s from _somewhere_, and only then the user could fetch logs.
* There was no straightforward, nor documented way to get `TaskInvocationId`s of interest.

# This PR's solution

This PR avoids the problem altogether by simplifying the `sdk.WorkflowRun.get_logs()` signature. Now, we no longer require the user to pass any IDs, we just return everything we've got. When the user is interested in logs for only a subset of tasks, they can either get the item by filtering the output on their own:
```python

wf_run = sdk.WorkflowRun.by_id("foo")

logs = wf_run.get_logs()
# Option 1
single_task_logs = logs["my_inv_id"]

# Option 2
logs_subset = {id: lines for id, lines in logs.items() if id in ["foo", "bar", "baz"]}

# Option 3
for task in wf_run.get_tasks():
    if task.get_status() == State.FAILED:
        print(task.get_logs())
```

Bonus points:
* we no longer have to worry about users passing invalid IDs
* we no longer have to document what kind of input we expect, and what are the implications passing `tasks=[]`, `tasks=["invalid_id", "valid_id"]`, etc.

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/cli/_dorq/_repos.py
+++ b/src/orquestra/sdk/_base/cli/_dorq/_repos.py
@@ -335,15 +335,15 @@ class WorkflowRunRepo:
     ) -> t.Mapping[TaskInvocationId, t.Sequence[str]]:
         """
         Raises:
-            orquestra.sdk.exceptions.NotFoundError: when the wf_run_id doesn't match a
-                stored run ID.
-            orquestra.sdk.exceptions.ConfigNameNotFoundError: when the named config is
-                not found in the file.
+            orquestra.sdk.exceptions.WorkflowRunNotFoundError
+            orquestra.sdk.exceptions.ConfigmeNotFoundError
+            orquestra.sdk.exceptions.ConfigNameNotFoundError
         """
         try:
             wf_run = sdk.WorkflowRun.by_id(wf_run_id, config_name)
         except (
             exceptions.NotFoundError,
+            exceptions.ConfigFileNotFoundError,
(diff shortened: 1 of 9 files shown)


src/orquestra/sdk/_base/cli/_dorq/_repos.py
```diff
@@ -306,47 +305,65 @@ def get_task_inv_ids(
...
-            logs = wf_run.get_logs(tasks=task_inv_id)
-        except (exceptions.WorkflowRunNotStarted, exceptions.TaskRunNotFound):
-            raise
+            task_run: sdk.TaskRun = _find_first(
+                lambda task: task.task_invocation_id == task_inv_id, task_runs
+            )
+        except StopIteration as e:
+            raise exceptions.TaskInvocationNotFoundError(task_inv_id) from e
+
+        # TODO: handle and document exceptions
```
SebastianMorawiec, line 359: Issue: hanging TODO

SebastianMorawiec (approved):
Approving - just do somethign with that TODO 👍

src/orquestra/sdk/_base/cli/_dorq/_repos.py
alexjuda: Good catch!

PR #33, opened 2023-01-26
fix: Return a single result (as expected) from CERuntime, in a tuple

jamesclark-Zapata:
# The problem

The CERuntime made an incorrect assumption about the shape of results.

It expected this workflow:

```python
@sdk.workflow
def wf():
    return 1,2,3
```

to return 3 workflow result IDs via the API.

In fact, it returned a single result ID.

This meant, `WorkflowRun.get_results()` could return `[[1,2,3]]` which looks odd.

# This PR's solution

 - Makes an assumption that there is one result per workflow run. This assumption is backed up by an assertion as a guard.
 - Also turns the JSON lists to tuples. Note: this assumes RayRuntime always returns tuples

# Checklist


alexjuda (approved):
LGTM as a temporary solution

PR #35, opened 2023-01-30
feat: GitImports in RayRuntime

jamesclark-Zapata:
# The problem

If users want to use GitImports inside the Ray on cloud, we need to support Git repos inside the Ray runtime.

# This PR's solution

Uses the `pip` section of Ray's `RuntimeEnv` to handle Git imports.
This only works for Git repos that are also Python packages.

# Checklist

### DIFF
--- a/src/orquestra/sdk/_ray/_dag.py
+++ b/src/orquestra/sdk/_ray/_dag.py
@@ -271,7 +271,7 @@ def _import_pip_env(ir_invocation: ir.TaskInvocation, wf: ir.WorkflowDef):
             *(task_def.dependency_import_ids or []),
         )
     ]
-    return list(itertools.chain(*(_pip_string(imp) for imp in imports)))
+    return [chunk for imp in imports for chunk in _pip_string(imp)]
 
 
 def _gather_args(
--- a/src/orquestra/sdk/_ray/_dag.py
+++ b/src/orquestra/sdk/_ray/_dag.py
@@ -8,12 +8,14 @@ RuntimeInterface implementation that uses Ray DAG/Ray Core API.
 from __future__ import annotations
 
 import dataclasses
+import itertools
 import json
 import logging
 import re
 import traceback
 import typing as t
 from datetime import datetime, timedelta, timezone
+from functools import singledispatch
 from pathlib import Path
 
 from orquestra.sdk import exceptions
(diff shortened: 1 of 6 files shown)


src/orquestra/sdk/_ray/_dag.py
```diff
@@ -238,16 +240,38 @@ class WfUserMetadata(t.TypedDict):
...
+    task_def = wf.tasks[ir_invocation.task_id]
+    imports = [
+        wf.imports[id_]
+        for id_ in (
+            task_def.source_import_id,
+            *(task_def.dependency_import_ids or []),
+        )
     ]
-    return [package for imp in python_imports for package in imp.packages]
+    return list(itertools.chain(*(_pip_string(imp) for imp in imports)))
```
alexjuda, line 274: **non-blocking**: nested list comprehension would be easier for me to interpret than `list()` + `chain()` + `*`

```suggestion
    return [chunk for imp in imports for chunk in _pip_string(imp)]
```

src/orquestra/sdk/_ray/_dag.py
```diff
@@ -238,16 +240,38 @@ class WfUserMetadata(t.TypedDict):
...
+@_pip_string.register
+def _(imp: ir.GitImport):
+    m = re.match(
+        r"(?P<user>.+)@(?P<domain>[^/]+?):(?P<repo>.+)", imp.repo_url, re.IGNORECASE
+    )
+    if m is not None:
+        url = f"ssh://{m.group('user')}@{m.group('domain')}/{m.group('repo')}"
+    else:
+        url = imp.repo_url
+    return [f"git+{url}@{imp.git_ref}"]
```
alexjuda, line 262: **question**: I've seen you added an integration test for it. The coverage report haven't recorded testing this. Is it because this code is executed in another process?

alexjuda: **suggestion**: could you also add a unit test for this function? Mistakes in regexps are easy to miss, and unit testing regexp-based code is usually easy to set up.

alexjuda (approved):
Added non-blocking comments

src/orquestra/sdk/_ray/_dag.py
jamesclark-Zapata: @alexjuda yeah, this is executed inside Ray which doesn't get picked up by the coverage tool.
I'll add some unit tests for `_pip_string`

PR #38, opened 2023-02-01
fix: return only structured logs

alexjuda:
# The problem

Logs read from workflows submitted to Ray resulted in a mix of domains and formats:
```
2023-01-02 09:50:58,793 INFO task_executor.py:78 -- Task status [RUNNING]   [wf.regression_demo.8f92f5f@invocation-1-task-train-model]
{"timestamp": "2023-01-02 09:50:58,809", "level": "ERROR", "filename": "_dag.py:194", "message": {"run_id": "wf.regression_demo.8f92f5f@invocation... 
2023-01-02 09:55:43,523 INFO task_executor.py:78 -- Task status [RUNNING]   [wf.broken_regression_demo.6014772@invocation-1-task-train-model]
```

You can see here both Ray-produced logs in plain-text and user-generated logs in JSONL.

# This PR's solution

* Reading logs now only retrieves user-generated lines.
* To make sure this PR's solution works I've ended up reworking how we generate logs, too.
    * Bonus: we no longer set the global logging config. Fixes https://zapatacomputing.atlassian.net/browse/ORQSDK-740.
    * Bonus: we're using now ISO-8601 date strings with timezones. Fixes https://zapatacomputing.atlassian.net/browse/ORQSDK-563

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_log_adapter.py
+++ b/src/orquestra/sdk/_base/_log_adapter.py
@@ -8,14 +8,15 @@ Log adapter adds a workflow context to logs, Workflow ID and Task ID.
 import json
 import logging
 import os
-from typing import Tuple
+import typing as t
+from datetime import datetime, timezone
 
 from orquestra.sdk.schema.workflow_run import TaskRunId, WorkflowRunId
 
-RAY_LOGS_FORMAT = '{"timestamp": "%(asctime)s", "level": "%(levelname)s", "filename": "%(filename)s:%(lineno)s", "message": %(message)s}'  # noqa
-logging.basicConfig(
-    format=RAY_LOGS_FORMAT,
-)
+# NOTE: the `message`, `wf_run_id`, and `task_run_id` value placeholders don't come with
+# "" quotes. We already add them when we json.dumps() the value. This ensure proper JSON
+# escaping and handling null values.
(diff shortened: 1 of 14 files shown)


src/orquestra/sdk/_base/_log_adapter.py
```diff
@@ -8,31 +8,45 @@
...
+    ) -> t.Tuple[str, t.MutableMapping[str, t.Any]]:
+        """
+        Args:
+            msg: message passed to the standard logger method calls.
+            kwargs: custom kwargs passed to the standard logger method calls.
+        """
+
+        # Ensure the message can be stored as a JSON. This will escape any
+        # JSON-forbidden chars.
+        new_msg = json.dumps(msg)
```
jamesclark-Zapata, line 38: **question:** ‏I assume msg is always guaranteed to be a string because it comes after any `f"{f_string}"` or `"formatted string {}".format(object())`?

jamesclark-Zapata (approved):

src/orquestra/sdk/_base/_log_adapter.py
alexjuda: Good catch! I'll wrap it in `str()` so we don't have to guess.

PR #39, opened 2023-02-03
chore: Replace config_save_file with an environment variable

jamesclark-Zapata:
# The problem

It was difficult to maintain the code and tests with passing around the config path around.

# This PR's solution

Removes the `config_save_file` option and replaces with an env variable.

# Checklist

### DIFF
--- a/tests/sdk/v2/test_config.py
+++ b/tests/sdk/v2/test_config.py
@@ -18,6 +18,9 @@ the user. It's a lot easier to figure out appropriate behavior this way.
 """
 import datetime
 import json
+import os
+import tempfile
+from pathlib import Path
 from unittest.mock import Mock
 
 import filelock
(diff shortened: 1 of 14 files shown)


tests/sdk/v2/test_config.py
```diff
@@ -29,6 +32,13 @@
 
 from .data.configs import TEST_CONFIG_JSON
 
+# @pytest.fixture(scope="module", autouse=True)
+# def mock_config_path():
+#     with tempfile.TemporaryDirectory() as dir_path:
+#         with pytest.MonkeyPatch.context() as mp:
+#             mp.setenv("ORQ_CONFIG_PATH", os.path.join(dir_path, "config.json"))
+#             yield mp
+
```
alexjuda, line 41: Leftover?
```suggestion
```

alexjuda (approved):
LGTM!

**non-blocking,thought**: it would be nice to have a single place where we document all the env variables we use. Do you think it would make a good docs page in the future?

PR #40, opened 2023-02-03
chore!: Switch corq to dorq

alexjuda:
https://zapatacomputing.atlassian.net/browse/ORQSDK-748

# The problem

Users won't use our new CLI unless we expose it under a nice alias.

# This PR's solution

* Point `orq` to dorq CLI
* Fix CLI argument wording
* Adopt CLI reference doc for click

# Checklist


tests/runtime/performance/test_cli_perf.py
```diff
@@ -11,8 +11,10 @@
 https://zapatacomputing.atlassian.net/browse/ORQSDK-507
```
jamesclark-Zapata, line 11: **question:** ‏Are these relevant now? Should they be switched to the new CLI?

jamesclark-Zapata (approved):
question about perf test, but otherwise lgtm

jamesclark-Zapata (approved):
question about perf test, but otherwise lgtm

tests/runtime/performance/test_cli_perf.py
alexjuda: They do test that corq still works. I think we should keep them as long as we keep corq code.

alexjuda: > Should they be switched to the new CLI?

Having similar ones for dorq sounds like a good idea, but I'd keep it out of scope. Added under https://zapatacomputing.atlassian.net/browse/ORQSDK-749

PR #51, opened 2023-02-09
fix: reading single-task logs on Ray

alexjuda:
Marked as "draft" until the dependent PRs are merged:
* https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/49
* https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/50

# The problem

`orq task logs` (and the corresponding Python API) didn't work with reading task logs from Ray.

# This PR's solution

* Splits the awkward `RuntimeInterface.get_full_logs()` into to separate methods: `.get_workflow_logs()` and `.get_task_logs()` with non-optional arguments.
* Reimplements the log reading for Ray logs.
* Simplifies the log reading implementation for QE.

Main ticket: https://zapatacomputing.atlassian.net/browse/ORQSDK-678. Also solves:
- https://zapatacomputing.atlassian.net/browse/ORQSDK-750
- https://zapatacomputing.atlassian.net/browse/ORQSDK-584

# Checklist


src/orquestra/sdk/_ray/_ray_logs.py
```diff
@@ -125,34 +121,33 @@ def _update_log_filenames(self):
...
+
+                    file_info.file_position = f.tell()
 
-    def get_full_logs(self):
-        parsed_logs = self._read_log_files()
-        formatted = [log.json() for log in parsed_logs]
-        # This is bad. The key should be a task invocation ID. To be fixed in the Jira
-        # ticket: https://zapatacomputing.atlassian.net/browse/ORQSDK-676
-        return {"logs": formatted}
+        return log_lines
```
jamesclark-Zapata, line 150: **question:** ‏Can you explain why we return and yield? Whats the example usecase?

alexjuda: You've found a leftover after the prev implementation, good catch! I'll update the PR, meanwhile see the motivation below.

The first approach from this PR read all lines from all files, returned them as a one big list, and the lines where being filtered only later on on another function.

The updated approach is a lazy sequence implemented as generator that yields line by line so don't have to keep big lists in memory only to dispose most of them. It's a memory optimization based on gut feeling. This is also similar to the implementation that was there before this PR, so I didn't want to make the performance terribly worse.

PR #52, opened 2023-02-09
feat: Allow secrets in workflow functions and IR

jamesclark-Zapata:
# The problem

Currently, if someone wanted to use a secret inside their workflow, they may accidentally expose the secret value as a constant inside the workflow graph.

# This PR's solution

- Adds a new exec context for workflow building
- Returns a "Secret" future for any call to `sdk.secrets.get` inside this new context
- Adds a new node to the IR for secrets
- Updates the traversal code to accept this new node

Note: the traversal code has changed, meaning the ordering of invocations, constants, and artifacts may have changed.
This does not impact any previously generated workflows.

Following PRs will add this functionality to:
- Ray
- QE
- In-process

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_dsl.py
+++ b/src/orquestra/sdk/_base/_dsl.py
@@ -73,7 +73,12 @@ class UnknownPlaceholderInCustomNameWarning(Warning):
 # ----- data structures -----
 
 Constant = Any
-Argument = Union[Constant, "ArtifactFuture"]
+Argument = Union[Constant, "ArtifactFuture", "Secret"]
+
+
+class Secret(NamedTuple):
+    name: str
+    config_name: Optional[str] = None
 
 
 class GitImport(NamedTuple):
--- a/src/orquestra/sdk/_base/_dsl.py
+++ b/src/orquestra/sdk/_base/_dsl.py
@@ -78,6 +78,10 @@ Argument = Union[Constant, "ArtifactFuture", "Secret"]
 
 class Secret(NamedTuple):
     name: str
+    # Config name is only used for the local runtimes where we can't infer the location
+    # where we get a secret's value from.
+    # This matches the behaviour of `sdk.secrets.get` where the config name is used to
+    # get a secret when running locally.
     config_name: Optional[str] = None
 
 
--- a/src/orquestra/sdk/_base/_traversal.py
+++ b/src/orquestra/sdk/_base/_traversal.py
@@ -90,32 +90,32 @@ class GraphTraversal:
         We iterate over the futures returned from the workflow find the artifacts,
         constants, and secrets.
         """
-        node_counts: t.Dict[str, int] = defaultdict(int)
+        artifact_counter = 0
+        secret_counter = 0
+        constant_counter = 0
         for n in _iter_nodes(root_futures):
-            node_type: str
             if isinstance(n, _dsl.ArtifactFuture):
-                node_type = "artifact"
                 self._artifacts[_make_key(n)] = _make_artifact_node(
-                    node_counts[node_type], n
+                    artifact_counter, n
                 )
+                artifact_counter += 1
                 # Map the invocation to the future.
                 # Note: Each unique future has one invocation, but each invocation
                 #       can have many Futures.
                 #       We're mapping `invocation: set(futures from invocation)`
                 self._invocations.setdefault(n.invocation, set()).add(n)
             elif isinstance(n, _dsl.Secret):
-                node_type = "secret"
                 self._secrets[_make_key(n)] = model.SecretNode(
-                    id=f"secret-{node_counts[node_type]}",
+                    id=f"secret-{secret_counter}",
                     secret_name=n.name,
                     secret_config=n.config_name,
                 )
+                secret_counter += 1
             else:
-                node_type = "constant"
                 self._constants[_make_key(n)] = _make_constant_node(
-                    node_counts[node_type], n
+                    constant_counter, n
                 )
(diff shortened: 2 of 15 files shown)


src/orquestra/sdk/_base/_dsl.py
```diff
@@ -73,7 +73,12 @@ class UnknownPlaceholderInCustomNameWarning(Warning):
 # ----- data structures -----
 
 Constant = Any
-Argument = Union[Constant, "ArtifactFuture"]
+Argument = Union[Constant, "ArtifactFuture", "Secret"]
+
+
+class Secret(NamedTuple):
+    name: str
+    config_name: Optional[str] = None
```
alexjuda, line 81: **non-blocking**: please add a comment describing when we expect this field to be `None`

src/orquestra/sdk/_base/_traversal.py
```diff
@@ -73,26 +73,86 @@ def _make_artifact_id(source_task: model.TaskDef, future_index: int):
...
+        self._constants = {}
+
+    def traverse(self, root_futures: t.Sequence[GraphNode]):
+        """
+        Traverse the workflow graph.
+
+        We iterate over the futures returned from the workflow find the artifacts,
+        constants, and secrets.
+        """
+        node_counts: t.Dict[str, int] = defaultdict(int)
```
alexjuda, line 93: **non-blocking,suggestion**: looks like there's a well-known number of entries in `node_counts`. It would make it much more readable for me if this code used just three counter variables named explicitly:
```python
artifact_counter = 0
secret_counter = 0
constant_counter = 0
```
and then:
```python
self._artifacts[_make_key(n)] = _make_artifact_node(artifact_counter, n)
```
and:
```python
artifact_counter += 1
```

src/orquestra/sdk/_base/_workflow.py
```diff
@@ -256,6 +259,13 @@ def __call__(self, *args: _P.args, **kwargs: _P.kwargs) -> WorkflowDef[_R]:
             orquestra.sdk.exceptions.WorkflowSyntaxError: if the arguments don't match
                 the workflow def's parameters.
         """
+        allowed_function_calls = (
+            Secret,
+            secrets.get,
+            secrets.delete,
+            secrets.list,
+            secrets.set,
```
alexjuda, line 267: **question**: is there any reason to call `secrets.{delete,list,set}` inside a workflow fn?

src/orquestra/sdk/schema/ir.py
```diff
@@ -269,7 +270,22 @@ class ConstantNodePickle(BaseModel):
...
+    A reference to a secret stored in an external secret/config service.
+    """
+
+    # Workflow-scope unique ID used to refer from task invocations
+    id: SecretNodeId
+
+    # Serialized value
+    secret_name: str
+
+    # Secret config
```
alexjuda, line 284: **non-blocking,suggestion**: please state in the comment what's a situation where we expect this to be `None`

src/orquestra/sdk/secrets/_api.py
```diff
@@ -46,6 +46,8 @@ def get(
         orquestra.sdk.exceptions.UnauthorizedError: when the authorization with the
             remote vault failed.
     """
+    if _exec_ctx.global_context == _exec_ctx.ExecContext.WORKFLOW_BUILD:
+        return t.cast(str, _dsl.Secret(name=name, config_name=config_name))
```
alexjuda, line 50: **issue**: returning `_dsl.Secret` "future" is inconsistent with the current docstring. I know this is what we want for the DSL purposes; please mention how the behavior of this function changes depending on the place where its used.

alexjuda (changes requested):
Added few non-blocking suggestions and one issue. Marking as blocked to make sure don't forget about the docstring!

src/orquestra/sdk/_base/_workflow.py
jamesclark-Zapata: No, but it seemed weird to me to allow getting secrets and none of the other bits of the API

alexjuda: Now, if I use `secrets.list()` it would produce a submit-time constant, which is very different to `secrets.get()` which produces a Secret node.

src/orquestra/sdk/secrets/_api.py
jamesclark-Zapata: wdyt? d0438b9

src/orquestra/sdk/schema/ir.py
jamesclark-Zapata: wdyt? 543cc88

src/orquestra/sdk/_base/_dsl.py
jamesclark-Zapata: wdyt? 543cc88

src/orquestra/sdk/_base/_workflow.py
jamesclark-Zapata: This wouldn't embed any secret values though.

This change here only related to if the `NotATaskWarning` is printed.

src/orquestra/sdk/_base/_dsl.py
alexjuda: Nice!

src/orquestra/sdk/_base/_workflow.py
alexjuda: Fair enough!

alexjuda (approved):

PR #53, opened 2023-02-09
feat: IR Secrets in runtimes

jamesclark-Zapata:
# The problem

Secrets are in the IR (#52) and now we need to implement runtime support

# This PR's solution

- Update Ray, QE, and in-process runtimes to get the secrets from the secrets API

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_conversions/_invocations.py
+++ b/src/orquestra/sdk/_base/_conversions/_invocations.py
@@ -10,6 +10,7 @@ from orquestra.sdk.schema import ir, responses, yaml_model
 # TODO: it would be nice to have a single document where the user could look up a "type"
 # from the workflow yaml and see what it's about.
 RESULT_DICT_TYPE_NAME = "workflow-result-dict"
+SECRET_DICT_TYPE_NAME = "workflow-secret-dict"
 SDK_METADATA_TYPE_NAME = "sdk-metadata"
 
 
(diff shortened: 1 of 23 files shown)


src/orquestra/sdk/_base/_conversions/_invocations.py
```diff
@@ -311,6 +314,14 @@ def _make_constant_input(
     }
 
 
+def _make_secret_input(arg_name: str, secret: ir.SecretNode) -> yaml_model.StepInput:
+    json_dict = json.loads(secret.json())
```
alexjuda, line 318: **thought**: it would be nice to have a single place where we transform a Pydantic to a JSON-compatible dict instead of `json.loads(model.json())` all over the place. 

For example, is there a reason to include `null`s in the output YAML? The default behavior of `BaseModel.json()` is to include `None`s

src/orquestra/sdk/_base/_in_process_runtime.py
```diff
@@ -95,6 +96,10 @@ def create_workflow_run(self, workflow_def: ir.WorkflowDef) -> WfRunId:
             id: deserialize_constant(node)
             for id, node in workflow_def.constant_nodes.items()
         }
+        for id, secret in workflow_def.secret_nodes.items():
+            consts[id] = secrets.get(
```
alexjuda, line 100: **thought**: I like the conciseness of dumping secrets as constants instead of a separate collection, but let's keep in mind this is prone to ID collision bugs if we ever generate `ConstantNodeID` and `SecretNodeID` from similar value domains.

alexjuda (approved):

src/orquestra/sdk/_base/_in_process_runtime.py
jamesclark-Zapata: Yeah, if we ever have ID collisions then this would cause issues. ID collisions would cause problems in other places too, I think.

src/orquestra/sdk/_base/_conversions/_invocations.py
jamesclark-Zapata: I added a helper for this. I left the nulls for now, because when we export to YAML we handle this. f8f309a

PR #61, opened 2023-02-28
chore: update Compute Engine error handling

jamesclark-Zapata:
# The problem

The API spec for Compute Engine was updated.

# This PR's solution

Adds the error handling for the newly documented endpoints.

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_driver/_exceptions.py
+++ b/src/orquestra/sdk/_base/_driver/_exceptions.py
@@ -52,14 +52,43 @@ class InvalidWorkflowDefID(Exception):
 class InvalidWorkflowRunID(Exception):
     """
     Raised when an invalid Workflow Run ID is sent to the Workflow Driver
-
-    Workflow Run IDs are expected to be UUIDs, if an ID is sent that cannot be
-    parsed as a UUID, this exception will be raised.
     """
 
     def __init__(self, workflow_run_id: str):
         self.workflow_run_id = workflow_run_id
         super().__init__(workflow_run_id)
+
+
+class InvalidWorkflowRunArtifactID(Exception):
+    """
+    Raised when an invalid Workflow Run Artifact ID is sent to the Workflow Driver
+
+    Workflow Run Artifact IDs are expected to be UUIDs, if an ID is sent that cannot be
+    parsed as a UUID, this exception will be raised.
+    """
+
+    def __init__(self, workflow_run_artifact_id: str):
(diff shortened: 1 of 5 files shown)


src/orquestra/sdk/_base/_driver/_exceptions.py
```diff
@@ -52,16 +52,45 @@ def __init__(self, workflow_def_id: str):
...
+    Workflow Run Result IDs are expected to be UUIDs, if an ID is sent that cannot be
+    parsed as a UUID, this exception will be raised.
+    """
+
+    def __init__(self, workflow_run_result_id: str):
+        self.workflow_run_result_id = workflow_run_result_id
+        super().__init__(workflow_run_result_id)
+
+
+class ForbiddenError(Exception):
```
alexjuda, line 88: There's also `UnauthorizedError`. Do we need to differentiate between the two cases on the user-facing layer?

tests/sdk/v2/driver/test_client.py
```diff
@@ -322,18 +333,29 @@ def test_sets_auth(
             def test_unauthorized(endpoint_mocker, client: DriverClient):
                 endpoint_mocker(
                     # Specified in:
-                    # https://github.com/zapatacomputing/workflow-driver/blob/2b353476d5b0161da31584533be208611a131bdc/openapi/src/resources/workflow-definition.yaml#L26
+                    # https://github.com/zapatacomputing/workflow-driver/blob/main/openapi/src/resources/workflow-definitions.yaml#L25
```
alexjuda, line 336: **suggestion**: can we use permalinks instead of branches? It's possible to go from a commit permalink to the latest revision, but it's much more difficult in the opposite order. The diff between `commit...main` also allows us to see if schema was changed since the time we wrote the client.

alexjuda (approved):
Accepted with some nit-picking

src/orquestra/sdk/_base/_driver/_exceptions.py
jamesclark-Zapata: `UnauthorizedError` is on the user facing level and `ForbiddenError` is an internal exception. On the next layer up `InvalidTokenError` and `ForbiddenError` are wrapped together under `UnauthorizedError`, I don't see an immediate need to have separate exception on the user level.

alexjuda: Ohhh right, nevermind. I thought this was `orquestra.sdk.exceptions` but it's not.

PR #64, opened 2023-03-02
Feat: bump ray to 2.3

SebastianMorawiec:
# The problem

_Please describe the motivation for this PR. Bonus points for code examples, error messages, and screenshots._

# This PR's solution

_Please describe the PR's content. Bonus points for code examples, error messages, screenshots._

# Checklist


setup.cfg
```diff
@@ -69,7 +69,7 @@ where = src
 
 [options.extras_require]
 ray =
-    ray[default]==2.0.1
+    ray[default]==2.3.0
     # Ray 2.0.0 requires pyarrow but it's missing from their requirements.
```
jamesclark-Zapata, line 73: **question:** ‏Can you check this too?

src/orquestra/sdk/_ray/_client.py
```diff
@@ -112,6 +113,21 @@ def run_dag_async(
         ):
             ray.workflow.run_async(dag_node, workflow_id=workflow_id, metadata=metadata)
 
+            # Give some time to ray to actually submit workflow
+            submission_timeout_in_sec = 10
```
jamesclark-Zapata, line 117: **question:** ‏Does this mean submitting to Ray could take up to 10 seconds?

src/orquestra/sdk/_ray/_client.py
```diff
@@ -112,6 +113,21 @@ def run_dag_async(
...
+            while True:
+                try:
+                    self.get_workflow_status(workflow_id)
+                    break
+                except workflow_exceptions.WorkflowNotFoundError:
+                    if retries * retry_time_in_sec >= submission_timeout_in_sec:
+                        raise
+                    else:
+                        time.sleep(retry_time_in_sec)
+                        retries += 1
```
jamesclark-Zapata, line 129: **question:** ‏Is it possible to add tests for these lines?

src/orquestra/sdk/_ray/_client.py
SebastianMorawiec: It always could. This is only the timer that says "if it takes LONGER than 10 seconds, error out". This wont extent the length of submission itself - only the length of command so it waits until workflow is actually submitted.

src/orquestra/sdk/_ray/_client.py
SebastianMorawiec: Good point - ill see what I can do here

setup.cfg
SebastianMorawiec: You mean pyarrow dependency?

SebastianMorawiec: Ok I can see they have this dependency on their list: 
https://github.com/ray-project/ray/blob/master/python/requirements.txt#L62
Im gonna remove ours (unless we use pyarrow somewhere in our project in different places)

jamesclark-Zapata: Check that this is included when you `pip install ray[default]`

SebastianMorawiec: Good point - so it doesnt.
According to this issue: https://github.com/ray-project/ray/issues/25952
We should be using ray[data] as dependency as well, but it brings things like pandas, which im not certain we want. I'd leave pyarrow here only.

tests/runtime/performance/test_cli_perf.py
```diff
@@ -85,18 +86,10 @@ def orq_workflow_run(ray_cluster, orq_project_dir):
...
-            "-d",
-            orq_project_dir,
-            "-c",
-            "local",
-        ]
-    )
+    # Wait for the workflow to finish to make sure that getting the results will succeed
+    wf = sdk.WorkflowRun.by_id(workflow_id)
+    wf.wait_until_finished()
+
```
jamesclark-Zapata, line 92: The purpose of these tests were to make sure the CLI didn't slow down.

Did you mean to only remove the `_run_corq_command` call instead of moving it to after the `wf.wait_until_finished()`?

jamesclark-Zapata: Ah, never mind. This is the fixture :)

jamesclark-Zapata (approved):

PR #67, opened 2023-03-03
feat: Private Github repo support

jamesclark-Zapata:
# The problem

Users should be able to pass credentials for private Git repos rather than these being configured on the cluster.

# This PR's solution

* Uses the secrets API to manage personal access tokens for Github imports

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_dsl.py
+++ b/src/orquestra/sdk/_base/_dsl.py
@@ -85,6 +85,19 @@ class Secret(NamedTuple):
     config_name: Optional[str] = None
 
 
+class GitImportWithAuth(NamedTuple):
+    """
+    A task import that uses a private Git repo
+
+    Please use a helper method, such as GithubImport, instead of this class
+    """
+
+    repo_url: str
+    git_ref: str
+    username: Optional[str]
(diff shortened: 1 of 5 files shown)


src/orquestra/sdk/_base/_dsl.py
```diff
@@ -109,9 +122,19 @@ def infer(
         return DeferredGitImport(local_repo_path, git_ref)
 
 
-def GithubImport(repo: str, git_ref: str = "main"):
+def GithubImport(
+    repo: str,
+    git_ref: str = "main",
+    username: Optional[str] = None,
+    personal_access_token: Optional[Secret] = None,
```
jamesclark-Zapata, line 129: I was thinking about adding a defensive check on this:
```
if not isinstance(personal_access_token, Secret):
    raise TypeError(f"A Github personal access token should be passed as an `sdk.Secret`, got {type(personal_access_token}")
```
but this would cause exceptions during imports because `GithubImport` is usually put inside decorators.

Type checkers should show this and later the graph traversal code will fail if anything but a `sdk.Secret` was passed.

alexjuda (approved):

PR #70, opened 2023-03-03
fix: Make Ray dirs before starting Ray

jamesclark-Zapata:
# The problem

Ray was failing to start after #66 because the directory it set may not exist

# This PR's solution

Ensures all Ray dirs are made before starting Ray

There's an issue with the test on CI, see ORQSDK-771.

# Checklist


alexjuda (approved):
**thought**: in the future, it would be nice to capture the `stdout` and `stderr` and pass it to an exception object if the subprocess fails. This way the user will have a chance to debug the error.

PR #71, opened 2023-03-03
fix: Switch to `cloudpickle` for pickling

jamesclark-Zapata:
# The problem

`dill` wasn't able to pickle `orquestra.sdk`.

# This PR's solution

- Adds `__init__.py` to make sure there are no namespaces in `orquestra.sdk`.
- Switch out to `cloudpickle` as the pickling library.
  - Update parts of the SDK which relied on `dill`.
- Bump to pickle protocol 4.
- Adds a test to ensure `orquestra.sdk` is picklable.

# Checklist


alexjuda (dismissed):
Do we need to add copyright headings to the empty files?

jamesclark-Zapata:
> Do we need to add copyright headings to the empty files?

Good shout

alexjuda:
Hold on I'm getting `python3.10/site-packages/dill/_dill.py:1705: PicklingWarning: Cannot locate reference to <class 'typing.ParamSpec'>.`

alexjuda:
My testing task:
```pythong
@sdk.task(source_import=sdk.InlineImport())
def sum_with_logs(a, b):
    """
    Example of a task with logging.
    """
    # Using this task causes serialization errors until
    # https://zapatacomputing.atlassian.net/browse/ORQSDK-770 is fixed.
    # TODO: add it to the testing workflow when the bug's fixed.

    logger = sdk.workflow_logger()

    sdk.wfprint(f"Adding two numbers: {a}, {b}")

    if a == b:
        logger.warning("Added numbers are the same! %s", a)

    return a + b
```

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 
 🚨 *Breaking Changes*
+* Pickling library switched to `cloudpickle` instead of `dill`.
```
alexjuda, line 7: **question**: What's the user impact, or why should a user care about this change?

setup.cfg
```diff
@@ -28,6 +28,8 @@ install_requires =
     # Schema definition 
     pydantic<2
     # Pickling library
+    cloudpickle==2.2.1
```
alexjuda, line 31: **question**: Is there a good reason to keep the requirement so strict?

src/orquestra/sdk/_base/_dsl.py
```diff
@@ -625,6 +625,9 @@ def __iter__(self):
         ]
         return iter(futures)
 
+    def __reduce_ex__(self, protocol):
+        raise NotImplementedError("ArtifactFuture cannot be pickled")
```
alexjuda, line 629: **question**: what happens if this method isn't defined at all?

src/orquestra/sdk/_base/serde.py
```diff
@@ -34,11 +35,28 @@ def _chunkify(s: str) -> t.List[str]:
...
     )
 
 
+@contextmanager
+def registered_module(module):
+    if module is not None:
+        register_pickle_by_value(module)
+    yield
+    if module is not None:
+        unregister_pickle_by_value(module)
```
alexjuda, line 49: **issue**: `unregister_` won't be called if `yield` throws an exception. [More info](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager).

```suggestion
    try:
        yield
    finally:
        if module is not None:
            unregister_pickle_by_value(module)
```

src/orquestra/sdk/_base/serde.py
```diff
@@ -34,11 +35,28 @@ def _chunkify(s: str) -> t.List[str]:
...
     return _chunkify(
         codecs.encode(
-            dill.dumps(object, protocol=PICKLE_PROTOCOL, recurse=True), ENCODING
+            cloudpickle.dumps(object, protocol=PICKLE_PROTOCOL), ENCODING
         ).decode()
     )
 
 
+@contextmanager
+def registered_module(module):
```
alexjuda, line 44: **suggestion**: could you add a docstring that states why we need it/when it's good to use it?

alexjuda (changes requested):
Nice job figuring this out! I've found just a few small things

setup.cfg
jamesclark-Zapata: We ran into an issue with `dill` where the internal changed between `0.3.4` and `0.3.6` and some environments where installing the old version for some reason.

Because of the change, some pickles were failing to be unpickled at runtime.

This requirement pin is a preemptive move to prevent this happening with `cloudpickle`.

src/orquestra/sdk/_base/_dsl.py
jamesclark-Zapata: If this isn't defined, `ArtifactFuture` objects can be pickled. This would happen when an artifact is inside a container, such as a `dict` or `list`.

We currently don't support this, so we need to prevent this from happening. With `dill`, it failed to pickle something else in this instance so it kinda worked by accident.

src/orquestra/sdk/_base/serde.py
jamesclark-Zapata: Good spot. You can also check out this [helpful blog post](https://jamesclark.dev/context-managers/#helper-decorator) for more info. :^)

src/orquestra/sdk/_base/serde.py
jamesclark-Zapata: Wdyt about 0edb446?

CHANGELOG.md
jamesclark-Zapata: Wdyt? 2fa99be

setup.cfg
alexjuda: Cloudpickle and dill are likely to be used by other packages our users might use alongside the SDK, eg scikit-learn. We need to be wary of dependency conflicts reported by users.

src/orquestra/sdk/_base/_dsl.py
alexjuda: **non-blocking**: please state that in a comment or docstring

alexjuda (approved):

src/orquestra/sdk/_base/_dsl.py
jamesclark-Zapata: done in 162fd65

PR #73, opened 2023-03-08
Chore: remove skipping service integration tests

SebastianMorawiec:
# The problem

We were unable to run the services integration tests due to the issues in PATHs - RAY executable couldnt be found.

# This PR's solution

Initial solution to just add RAY executable to path was problematic, because of the differences between OS's. To fix the problem in its roots, this PR properly creates and sources python venv, so the paths are correctly set and no more shenanigans in makefiles are needed.

# Checklist

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -36,6 +36,7 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
           architecture: x64
+      # Used to create OS-agnostic python virtual env used for other steps.
       - uses: syphar/restore-virtualenv@v1
         id: cache-virtualenv
 
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -32,21 +32,12 @@ jobs:
         with:
           # Fetch depth 0 required to compare against `main`
           fetch-depth: 0
-
-      # Load a specific version of Python
-      - name: Setup python
-        uses: actions/setup-python@v4
+      - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python }}
           architecture: x64
-
-      - name: Run Python commands
-        shell: bash
-        run: |
-          pip install --upgrade pip
-          python -m venv env
-          source env/bin/activate
-          echo "VIRTUAL ENV:" $VIRTUAL_ENV
+      - uses: syphar/restore-virtualenv@v1
+        id: cache-virtualenv
 
       - name: Install deps
         shell: bash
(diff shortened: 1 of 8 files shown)


.github/workflows/coverage.yml
```diff
@@ -32,13 +32,12 @@ jobs:
...
-
-      # Load a specific version of Python
-      - name: Setup python
-        uses: actions/setup-python@v4
+      - uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python }}
           architecture: x64
+      - uses: syphar/restore-virtualenv@v1
+        id: cache-virtualenv
```
jamesclark-Zapata, line 40: **suggestion:** ‏Can you add a comment for what this action is for/why it's needed? Same for everywhere it's used.

.github/workflows/publish_release.yml
```diff
@@ -18,13 +18,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      # Needed to get access to publish-release action definition
-      - name: Check out the released repo
-        uses: actions/checkout@v3
-        with:
-          # Fetch whole repo to get access to tags to read current package
-          # version.
-          fetch-depth: 0
```
jamesclark-Zapata, line 27: **issue:** ‏Did you mean to remove this? If so, will releases work?

Makefile
```diff
@@ -64,9 +64,8 @@ user-typing:
 
 # (override)
 github_actions:
-	${PYTHON_EXE} -m venv ${VENV_NAME}
-	${VENV_NAME}/${VENV_BINDIR}/${PYTHON_EXE} -m pip install --upgrade pip
-	${VENV_NAME}/${VENV_BINDIR}/${PYTHON_EXE} -m pip install -e '.[dev]'
+	${PYTHON} -m pip install --upgrade pip
+	${PYTHON} -m pip install -e '.[dev]'
```
jamesclark-Zapata, line 68: **nitpick (non-blocking):** ‏The other Makefile actions use `$(PYTHON)` syntax, there's no difference (iirc) but it would be nice to stick to one style.

.github/workflows/publish_release.yml
SebastianMorawiec: Good catch, bringing it back

tests/runtime/ray/test_integration.py
```diff
@@ -393,32 +393,6 @@ def test_after_a_failed_task(self, runtime: _dag.RayRuntime):
             inv_output = list(outputs.values())[0]
             assert inv_output == (58,)
 
-        def test_before_any_task_finishes(self, runtime: _dag.RayRuntime, tmp_path):
```
SebastianMorawiec, line 396: This test causes stucks due to the Ray not liking IO operations on Windows, and it tests pretty much redundant to other tests that we have

jamesclark-Zapata (approved):

PR #75, opened 2023-03-09
fix!: multi-output tasks when some outputs are unused

alexjuda:
# The problem

Sometimes, traversing workflow graph would produce invalid artifact nodes.

## Example

In the workflow below, using `two_outputs` should produce two artifacts, even if only one is used in the workflow. Otherwise, we get all sorts of problems, like missing values in `orq task results` or invalid nested tuples when the runtimes pack/unpack calculated task output values.

```python
from orquestra import sdk

@sdk.task
def two_outputs(a, b):
    return a + b, a - b

@sdk.workflow
def wf_with_multi_outputs():
    out1, _ = two_outputs(5, 1)
    return out1
```

# This PR's solution

1. Exposes `TaskOutputMetadata` in the IR schema. We've been using this data structure for some time on the DSL layer to represent the task def data shape, inferred statically. I needed to access this information in the CLI code, and the CLI code sees the workflow def IR, not the DSL.
2. Refactors `GraphTraversal`. Now, for a task def with $n$ outputs, the task invocation will point to $n + 1$ artifacts if the output is subscriptable or $1$ otherwise. See also the diagrams below.
3. Changes the contract for `RuntimeInterface.get_available_outputs()`. Now, in each entry in the dictionary the value is whatever object the task function returned. If the task returns an `int`, the entry's value is an `int`.
4. Removes `external_{file,module}_task`. It was never exposed in the public API. Implementing this PR's fix for it seemed like wasted effort.

# Examples

## Single output task

The simplest example.

```python
@sdk.task
def one_out(a, b):
    # We're returning a single number. This task's
    # inferred output metadata is subscriptable = False,
    # n_outputs = 1.
    return a + b

@sdk.workflow
def wf():
    out = single(5, 1)
    return out
```

```
      ┌─────────┐
      │inv-1-tsk│
      └────┬────┘
           ▼
      ┌────────────┐
      │art-1       │
      │index = None│
      └────────────┘
```

## Multi output task, both unpacked and not unpacked

The most complicated example.

```python
@sdk.task
def two_outs(a, b):
    # We're returning two numbers in a tuple. This task's
    # inferred output metadata is subscriptable = True,
    # n_outputs = 2.
    return a + b, a - b

@sdk.workflow
def wf():
    # not unpacked usage
    outs = two_outs(5, 1)

    # unpacking the tuple
    out1, out2 = outs

    return out1, out2, outs
```

```
                       ┌─────────┐
                       │inv-1-tsk│
                       └────┬────┘
              ┌──────────┬──┴────────┐
              ▼          ▼           ▼
      ┌─────────┐  ┌─────────┐  ┌────────────┐
      │art-1    │  │art-2    │  │art-3       │
      │index = 0│  │index = 1│  │index = None│
      └─────────┘  └─────────┘  └────────────┘
```

## Multi output task, unpacked

```python
@sdk.workflow
def wf():
    _, out2 = two_outs(5, 1)
    return out2
```

In this example, the workflow only makes use of artifact `art-2`, but we need
to keep track of the other artifacts anyway:
* Theoretically, we could drop the `ArtifactNode(id="art-1")` from the wf def graph, but it would make the code for graph traversal harder. Doing so would also complicate the "dispatcher" used on QE to store artifacts on disk when the task ends.
* Storing the unpacked the value for `art-3` is required to power getting intermediate values via `$ orq task results`/`wf_run.get_artifacts()`.

```
                       ┌─────────┐
                       │inv-1-tsk│
                       └────┬────┘
              ┌──────────┬──┴────────┐
              ▼          ▼           ▼
      ┌─────────┐  ┌─────────┐  ┌────────────┐
      │art-1    │  │art-2    │  │art-3       │
      │index = 0│  │index = 1│  │index = None│
      └─────────┘  └─────────┘  └────────────┘
```

# Checklist


BenjaminMummery (commented):
LGTM, a couple of spelling and stylistic suggestions. I really like the shift from `ir` to `model`, I think it makes the code more readable!

BenjaminMummery (approved):

jamesclark-Zapata (changes requested):
Would like to discuss this further w/ Alex

alexjuda:
Thanks for your review and comments, @BenjaminMummery!

@jamesclark-Zapata and I talked in another chat. I'm gonna spend some time figuring out how to make it all work without the rigid constraints imposed by this PR because of the following:
1. It goes against our goal of being as close to regular Python as possible. In regular Python you can unpack a tuple whenever you want, regardless of the function def.
2. Tasks defined in a 3rd-party library would impose the usage pattern without a way to change it on the workflow side.
3. We already have users whose workflows would break by introducing this PR.

src/orquestra/sdk/schema/ir.py
```diff
@@ -201,6 +217,10 @@ class TaskDef(BaseModel):
     # An empty list [] means a Task with no parameters
     parameters: t.Optional[t.List[TaskParameter]]
 
+    # Statically inferred from the task function. See also `TaskOutputMetadata`'s
+    # docstring.
+    output_metadata: TaskOutputMetadata
```
alexjuda, line 222: **thought**: this is a backwards incompatible change. It's likely that users won't be able to interact with old workflows after upgrading the SDK/CLI version.

However, any bugfix for this PR's motivating issue is likely to change how we process workflow graphs, so it would be backwards-incompatible anyway.

src/orquestra/sdk/_base/_traversal.py
```diff
@@ -72,43 +71,195 @@ def _gen_id_hash(*args):
...
+                # There can be many futures requiring to run the same task invocation.
+                if n.invocation in seen_invocations:
+                    # We've already handled the invocation, but not this future. We're
+                    # supposed to have artifact nodes already generated for it. We just
+                    # need to point "self._future_artifacts" to it.
+
+                    if out_meta.is_subscriptable:
+                        assert n.invocation in self._invocation_unpacked_artifacts
+
+                    assert n.invocation in self._invocation_non_unpacked_artifacts
```
jamesclark-Zapata, line 217: **question:** ‏Can you explain these asserts?

src/orquestra/sdk/_base/_traversal.py
```diff
@@ -72,43 +71,195 @@ def _gen_id_hash(*args):
...
+
+                    self._point_future_to_artifact(n)
+                else:
+                    # We haven't seen the invocation and haven't seen the future. We'll
+                    # have to generate artifact nodes it before we can point
```
jamesclark-Zapata, line 222:
```suggestion
                    # have to generate artifact nodes before we can point
```

src/orquestra/sdk/_base/_traversal.py
```diff
@@ -72,43 +71,195 @@ def _gen_id_hash(*args):
...
+                    # "self._future_artifacts" to it.
+
+                    if out_meta.is_subscriptable:
+                        # Generate artifact nodes for unpacking ("bar" and "baz" in the
+                        # example above).
+                        unpacked_artifacts = []
+                        for output_i in range(out_meta.n_outputs):
```
jamesclark-Zapata, line 229: **question:** ‏Do we now store all unpacked artifacts in the IR even if they aren't used?

jamesclark-Zapata (changes requested):
I think I'll need to take another look later, to make sure I understand what's going on.

In the meantime, I have some comments/nits.

src/orquestra/sdk/schema/ir.py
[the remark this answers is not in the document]
jamesclark-Zapata: I saw you commented on this after I started reviewing.

I think we'll need to be very explicit about this. We don't want to delay folks upgrading, and a breaking change may put folks off.
On the other hand, it might be better to rip the band-aid off now.

src/orquestra/sdk/_base/_api/_wf_run.py
```diff
@@ -383,13 +384,8 @@ def get_artifacts(self) -> t.Mapping[ir.TaskInvocationId, t.Any]:
...
         # getting a subset of artifacts, we should use them here.
-        workflow_artifacts = self._runtime.get_available_outputs(run_id)
-
-        # The runtime always returns tuples, even of the task has n_outputs = 1. We
-        # need to unwrap it for user's convenience.
-        return unwrap_task_retvals(
-            artifacts=workflow_artifacts, task_invocations=self._wf_def.task_invocations
-        )
+        inv_outputs = self._runtime.get_available_outputs(run_id)
+        return inv_outputs
```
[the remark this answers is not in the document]
jamesclark-Zapata: Note: the CE difference is because it isn't using this version of `orquestra-sdk`! It's using the last release.

src/orquestra/sdk/_base/_traversal.py
```diff
@@ -72,43 +71,195 @@ def _gen_id_hash(*args):
...
 
 
 class GraphTraversal:
     def __init__(self):
-        self._artifacts = {}
-        self._invocations = {}
-        self._secrets = {}
-        self._constants = {}
+        self._secrets: t.MutableMapping[t.Hashable, ir.SecretNode] = {}
+        self._constants: t.MutableMapping[t.Hashable, ir.ConstantNode] = {}
```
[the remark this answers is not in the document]
alexjuda: I disagree. Variable types should reflect the smallest interface surface needed, if possible. This is the same reason why it's preferable to write
```java
List<Integer> my_list = new ArrayList<Integer>();
```
than
```java
ArrayList<Integer> my_list = new ArrayList<Integer>();
```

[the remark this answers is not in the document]
jamesclark-Zapata: It's not so important for this PR, but my rule of thumb for typing is to return specifics but accept generics.

src/orquestra/sdk/_base/_in_process_runtime.py
```diff
@@ -155,14 +155,21 @@ def get_available_outputs(
 
         inv_outputs: t.Dict[ir.TaskInvocationId, TaskOutputs] = {}
         for inv in wf_def.task_invocations.values():
-            output_vals = tuple(
-                [
-                    self._artifact_store[workflow_run_id][art_id]
-                    for art_id in inv.output_ids
-                ]
-            )
+            # Assumption there's always a non-unpacked artifact. We want to return
```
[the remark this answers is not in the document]
alexjuda: This piece of code relies on how `GraphTraversal` produces artifact nodes; there's always at least a one artifact node related to the packed output value.

src/orquestra/sdk/schema/ir.py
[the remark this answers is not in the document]
alexjuda: Yeah, ripping band-aids sucks. Let's consider what other options we have:

1. Don't add `TaskDef.output_metadata` to the IR at all, and figure out some other way get this information in the CLI code (or rework CLI so it doesn't need this at all). 
    * Con 1: this still doesn't guard our users from bugs related to interacting with old workflow defs because I'm changing how `TaskInvocation.output_ids` are populated anyway (both on the producing and consuming side). 
    * Con 2: the CLI code might end up hacky, relying on even more implicit assumptions.
2. Add `TaskDef.output_metadata` to the IR, but make it optional. Old workflow defs get `None`, and the code that requires this information raises an exception telling the user this workflow is too old. `orq task results` is the only functionality I can think of; perhaps more can come in the future

src/orquestra/sdk/_base/_qe/_qe_runtime.py
```diff
@@ -342,6 +348,17 @@ def _http_error_handling():
...
 
 
+def _find_packed_artifact_id(
+    wf_def: WorkflowDef, inv_id: TaskInvocationId
+) -> ArtifactNodeId:
+    output_ids = wf_def.task_invocations[inv_id].output_ids
+    artifact_nodes = (wf_def.artifact_nodes[id] for id in output_ids)
+    packed: ArtifactNode = _find_first(
+        lambda node: node.artifact_index is None, artifact_nodes
+    )
```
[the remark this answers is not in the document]
jamesclark-Zapata: Can you add an assertion to guard against this?

src/orquestra/sdk/_base/_in_process_runtime.py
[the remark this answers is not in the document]
jamesclark-Zapata: Can you add an assertion to guard against this?

src/orquestra/sdk/_base/_traversal.py
jamesclark-Zapata: I think it was the two different assertions that I was confused about, but I understand now that there is always an artifact for the full return tuple and sometimes there's an unpacked one

CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
 
 🚨 *Breaking Changes*
 
-- Workflow definitions now require at least one task in order to be submitted. This check is performed during traversal, and raises a WorkflowSyntaxError if no tasks are required to be executed.
+* Pickling library switched to `cloudpickle` instead of `dill`. While no breakages are expected, this change may result in objects raising an error during pickling, even if they were previously able to be pickled. Please report any instances of these as bugs.
```
jamesclark-Zapata, line 7: This popped back up

CHANGELOG.md
```diff
@@ -4,7 +4,9 @@
...
+* Workflow definitions now require at least one task in order to be submitted. This check is performed during traversal, and raises a `WorkflowSyntaxError` if no tasks are required to be executed.
+* Workflow definitions now require at least one task in order to be submitted. This check is performed during traversal, and raises a WorkflowSyntaxError if no tasks are required to be executed.
```
jamesclark-Zapata, line 9: Duped

src/orquestra/sdk/_base/_in_process_runtime.py
```diff
@@ -161,11 +159,11 @@ def get_available_outputs(
...
             artifact_nodes = [wf_def.artifact_nodes[id] for id in inv.output_ids]
-            packed_artifact = [
-                artifact
-                for artifact in artifact_nodes
-                if artifact.artifact_index is None
-            ][0]
+            packed_nodes = [n for n in artifact_nodes if n.artifact_index is None]
+            assert len(
+                packed_nodes
+            ), f"Task invocation should have exactly 1 packed output. {inv.id} has {len(packed_nodes)}: {packed_nodes}"  # noqa: E501
```
jamesclark-Zapata, line 165: **issue:** ‏all non-zero values are truthy
```suggestion
            assert len(
                packed_nodes
            ) == 1, f"Task invocation should have exactly 1 packed output. {inv.id} has {len(packed_nodes)}: {packed_nodes}"  # noqa: E501
```

src/orquestra/sdk/_base/_qe/_qe_runtime.py
```diff
@@ -353,10 +352,12 @@ def _find_packed_artifact_id(
...
     output_ids = wf_def.task_invocations[inv_id].output_ids
     artifact_nodes = (wf_def.artifact_nodes[id] for id in output_ids)
-    packed: ArtifactNode = _find_first(
-        lambda node: node.artifact_index is None, artifact_nodes
-    )
-    return packed.id
+    packed_nodes = [n for n in artifact_nodes if n.artifact_index is None]
+    assert len(
+        packed_nodes
+    ), f"Task invocation should have exactly 1 packed output. {inv_id} has {len(packed_nodes)}: {packed_nodes}"  # noqa: E501
```
jamesclark-Zapata, line 358: **issue:** ‏all non-zero values are truthy
```suggestion
    assert len(
        packed_nodes
    ) == 1, f"Task invocation should have exactly 1 packed output. {inv_id} has {len(packed_nodes)}: {packed_nodes}"  # noqa: E501
```

CHANGELOG.md
```diff
@@ -14,7 +15,15 @@
...
-* Attempting to use the "in-process" runtime on the CLI will no longer raise an exception. Instead, a message teeling you to use the Python API or Ray will be printed.
+* Attempting to use the "in-process" runtime on the CLI will no longer raise an exception. Instead, a message telling you to use the Python API or Ray will be printed.
+* Fixed returning intermediate workflow values (e.g. with `orq task results`) when the task has multiple outputs and only some of them were used in the rest of the workflow function. The following should work now as expected:
+```python
+@sdk.workflow
+def my_wf():
+    _, b = two_output_task()
+    all_outputs = two_output_task()
+    out1, out2 = all_outputs
+    return b, all_outputs, out1, out2
```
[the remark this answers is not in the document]
jamesclark-Zapata: Please fix this!

src/orquestra/sdk/_base/_in_process_runtime.py
alexjuda: Fixed in cf600b6.

CHANGELOG.md
alexjuda: Fixed in 4ed1d1b

jamesclark-Zapata (approved):

PR #86, opened 2023-03-17
feat: raise an exception if workflows have zero tasks.

BenjaminMummery:
[ORQSDK-774]

# The problem

On QE you get an HTTP error stating that you need at least 1 task to be executed if your WF doesn't use any tasks. This is not the case on local Ray or CE. We should ensure that this is consistent across all runtimes.

# This PR's solution

- Adds a check during traversal that raises a WorkflowSyntaxError when no task invocations are identified. 
  - The exception reads: "The workflow '[wf name]' cannot be submitted as it does not define any tasks to be executed. Please modify the workflow definition to define at least one task and retry submitting the workflow."
  - If the source file and line can be identified, these are also included in the message.

# Checklist

[ORQSDK-774]: https://zapatacomputing.atlassian.net/browse/ORQSDK-774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/orquestra/sdk/_base/_workflow.py
+++ b/src/orquestra/sdk/_base/_workflow.py
@@ -118,13 +118,18 @@ class WorkflowDef(Generic[_R]):
         model = _traversal.flatten_graph(self, _futures)
 
         if len(model.task_invocations) < 1:
-            raise WorkflowSyntaxError(
-                f"The workflow '{model.name}' "
-                f"(defined at {model.fn_ref.file_path} line {model.fn_ref.line_number})"
-                " cannot be submitted as it does not define any tasks to be executed. "
+            helpstr = f"The workflow '{model.name}' "
+            if hasattr(model.fn_ref, "file_path"):  # If possible add the file and line.
+                helpstr += f"(defined at {model.fn_ref.file_path}"
+                if hasattr(model.fn_ref, "line_number"):
+                    helpstr += f" line {model.fn_ref.line_number}"
+                helpstr += ") "
+            helpstr += (
+                "cannot be submitted as it does not define any tasks to be executed. "
                 "Please modify the workflow definition to define at least one task "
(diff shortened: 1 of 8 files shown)


src/orquestra/sdk/_base/_workflow.py
```diff
@@ -113,7 +115,14 @@ def model(self) -> ir.WorkflowDef:
         else:
             _futures = futures
 
-        return _traversal.flatten_graph(self, _futures)
+        model = _traversal.flatten_graph(self, _futures)
+
+        if len(model.task_invocations) < 1:
+            raise WorkflowSyntaxError(
+                f"The workflow '{model.name}' requires 0 tasks to be executed. "
```
SebastianMorawiec, line 122: Suggestion: TBH I was a bit confused at first by this text. Maybe something more straight user-facing? Like "Your wofklow definition (or the workflow you are trying to submit? Whatever works better for you) [...]." 
But this thing here works for me as well.

SebastianMorawiec (approved):
👍

src/orquestra/sdk/_base/_workflow.py
BenjaminMummery: Good thought, thanks! [b383015](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/86/commits/b383015b8e90c27bae6045f6859dfd86de99141e) updates this to be a little more user-facing and also include the file and line number where the definition is found to help point the user to the changes they need to make. It now reads:

```
"The workflow '[wf name]' (defined at [file] line [line]) cannot be submitted as it does not define any tasks to be executed. Please modify the workflow definition to define at least one task and retry submitting the workflow."
```

PR #89, opened 2023-03-23
Fix: Move constant deserialization inside ray remote function

SebastianMorawiec:
# The problem

Constants were deserialised during dag creation. It causes that SDK would require all libraries required for all constants during dag creation time. It wasn't an issue in local ray, but is when using CE

# This PR's solution

Deserialise constants in ray.remote function with all proper task dependencies already installed in ray venv

# Checklist

### DIFF
--- a/src/orquestra/sdk/_ray/_dag.py
+++ b/src/orquestra/sdk/_ray/_dag.py
@@ -198,7 +198,7 @@ def _make_ray_dag_node(
     def _ray_remote(*inner_args, **inner_kwargs):
         if project_dir is not None:
             dispatch.ensure_sys_paths([str(project_dir)])
-        inner_args = (
+        inner_args = tuple(
             serde.deserialize_constant(arg)
             if isinstance(arg, t.get_args(ir.ConstantNode))
             else arg
--- a/src/orquestra/sdk/_ray/_dag.py
+++ b/src/orquestra/sdk/_ray/_dag.py
@@ -198,7 +198,18 @@ def _make_ray_dag_node(
     def _ray_remote(*inner_args, **inner_kwargs):
         if project_dir is not None:
             dispatch.ensure_sys_paths([str(project_dir)])
-
+        inner_args = (
+            serde.deserialize_constant(arg)
+            if isinstance(arg, t.get_args(ir.ConstantNode))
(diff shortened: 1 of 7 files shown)


src/orquestra/sdk/_ray/_dag.py
```diff
@@ -174,6 +174,8 @@ def _make_ray_dag_node(
     ray_kwargs: t.Mapping[str, t.Any],
     pos_unpack_specs: t.Sequence[PosArgUnpackSpec],
     kw_unpack_specs: t.Sequence[KwArgUnpackSpec],
+    pos_deserialize_specs: t.Sequence[int],
+    kw_deserialize_specs: t.Sequence[str],
```
alexjuda, line 178: **non-blocking**: `pos_deserialize_specs` is confusing to me because it's very similar to `pos_unpack_specs` but it works differently.
* `pos_arg_indices_to_deserialize`?
* `pos_deserialize_specs: t.Sequence[PosDeserializeSpec]` with a separate `PosDeserializeSpec` data structure similar to `PosArgUnpackSpec`?

alexjuda (approved):
LGTM! Added non-blocking suggestions :)

tests/runtime/ray/test_integration.py
```diff
@@ -545,18 +546,39 @@ def test_run_and_get_output(
...
+    them to task. This checks that we don't need python package outside an actual
+    task venv to deal with package-dependent constant values
+
+    We test those 2 things together as package installation is long, and we don't
+    want to do that twice unnecessarily
+    """
     # This package should not be installed before running test
     with pytest.raises(ModuleNotFoundError):
-        import piccup  # type: ignore # noqa
+        import polars  # type: ignore # noqa
```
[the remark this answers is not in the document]
SebastianMorawiec: dw, we still have 2nd test that imports that! :D

src/orquestra/sdk/_ray/_dag.py
SebastianMorawiec: I think creating separate type would be an overkill here - PosArgUpackSpec contains more info, we got just simple type, int or str. But will apply name change and doc it to be more explicit

PR #91, opened 2023-03-27
Feat: Allow setting default source imports and dependency imports for all tasks in given WF

SebastianMorawiec:
# The problem

Usually a lot of tasks in a single WF require the same dependencies. To this day users would have to define them separately in each and every task definition

# This PR's solution

User can define deafult-dependncies and default-source-import on WF level, which are used by each and every task inside a workflow unless task specifies their own dependencies. Then task-ones are overwriting default ones.

# Checklist


src/orquestra/sdk/_base/_dsl.py
```diff
@@ -397,6 +397,7 @@ def __init__(
         resources: Resources = Resources(),
         custom_image: Optional[str] = None,
         custom_name: Optional[str] = None,
+        custom_source_import: Optional[bool] = True,
```
alexjuda, line 400: Have you considered making `source_import` optional instead of adding a separate flag? 

Why I think it would be better: `custom_source_import` has a similar name to `custom_image` and `custom_name` but the `*_import`'s semantics is different; it's a flag that determines if the `source_import` should be used or ignored. Seems like a use case for optionals.

src/orquestra/sdk/_base/_workflow.py
```diff
@@ -523,6 +544,8 @@ def workflow(
     *,
     data_aggregation: Optional[Union[DataAggregation, bool]] = None,
     custom_name: Optional[str] = None,
+    default_source_import: Optional[Import] = None,
+    default_dependency_imports: Optional[Iterable[Import]] = None,
```
alexjuda, line 548: **issue**: please describe in the docstring (or on the docs( what's the implication of using vs not using `default_source_import` and `default_dependency_imports`. The `@sdk.workflow` decorator is the main component of this library, from all symbols we export this one should have the best documentation.

alexjuda (changes requested):

PR #92, opened 2023-03-27
refactor: remove obsolete RuntimeInterface methods

alexjuda:
# The problem

It's been a while since we reflected on `RuntimeInterface` methods. Originally this interface was public so we couldn't remove old methods; we only kept adding new ones. Nowadays, this interface isn't used by our users, so we're free to change it how we like.

# This PR's solution

Removes methods we don't use/implement anymore.

- `RuntimeInterface.get_workflow_run_outputs()`
- `RuntimeInterface.get_all_workflow_runs_status()`. The only usage is replaced with `list_workflow_runs()`.

# Checklist

### DIFF
--- a/src/orquestra/sdk/_ray/_dag.py
+++ b/src/orquestra/sdk/_ray/_dag.py
@@ -835,19 +835,21 @@ class RayRuntime(RuntimeInterface):
     def get_workflow_run_outputs_non_blocking(
         self, workflow_run_id: WorkflowRunId
     ) -> t.Sequence[t.Any]:
-        workflow_status = self.get_workflow_run_status(workflow_run_id)
+        try:
+            workflow_status = self.get_workflow_run_status(workflow_run_id)
+        except exceptions.WorkflowRunNotFoundError:
+            raise
+
         workflow_state = workflow_status.status.state
         if workflow_state != State.SUCCEEDED:
             raise exceptions.WorkflowRunNotSucceeded(
                 f"{workflow_run_id} has not succeeded. {workflow_state}",
                 workflow_state,
             )
-        try:
-            return self._client.get_workflow_output(workflow_run_id)
-        except ValueError as e:
-            raise exceptions.NotFoundError(
-                f"Workflow run {workflow_run_id} wasn't found"
-            ) from e
+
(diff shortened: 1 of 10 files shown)


src/orquestra/sdk/_ray/_dag.py
```diff
@@ -862,7 +842,12 @@ def get_workflow_run_outputs_non_blocking(
                 f"{workflow_run_id} has not succeeded. {workflow_state}",
                 workflow_state,
             )
-        return self.get_workflow_run_outputs(workflow_run_id)
+        try:
+            return self._client.get_workflow_output(workflow_run_id)
+        except ValueError as e:
+            raise exceptions.NotFoundError(
+                f"Workflow run {workflow_run_id} wasn't found"
+            ) from e
```
jamesclark-Zapata, line 850: **issue:** ‏These are uncovered, can you add a test to make sure this happens?

tests/runtime/ray/test_integration.py
```diff
@@ -546,81 +547,89 @@ def test_run_and_get_output(
...
-
-    We test those 2 things together as package installation is long, and we don't
-    want to do that twice unnecessarily
-    """
-    # This package should not be installed before running test
-    with pytest.raises(ModuleNotFoundError):
-        import polars  # type: ignore # noqa
+class Test3rdPartyLibraries:
+    @staticmethod
+    def test_constants_n_inline_imports(runtime: _dag.RayRuntime):
```
jamesclark-Zapata, line 573: **nitpick:** ‏Please update this test name, it looks incomplete.

alexjuda: What name would make more sense to you?

src/orquestra/sdk/_ray/_dag.py
alexjuda: It was a case that's unlikely to happen in normal circumstances. I simplified it in bad2622

tests/runtime/ray/test_integration.py
alexjuda: Reworded in 34627d5

jamesclark-Zapata (approved):

PR #94, opened 2023-03-28
internal: skip slow tests

alexjuda:
# Context

We have a mix of unit and integration tests, and the boundary is difficult to draw. Some tests take a long time to run. On the other hand, local development is very iterative, and we don't want to waste time re-running long tests unless necessary. This is why we have `make test-fast`.

# The problem

Some new tests have been added that slow `make test-fast` down.

# This PR's solution

1. I've marked any test case that takes more than 0.5s on my machine as "slow". This skips it from `make test-fast`. This doesn't affect the CI, it runs all tests. Of course, the threshold is totally based on a personal preference.
2. I've noticed we don't use some marks anymore, so I removed their definitions from `pytest.ini`.

Now, `make test-fast` takes "only" 20.59s. Before: 47.22s

# Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,6 @@ include subtrees/z_quantum_actions/Makefile
 # (override)
 test:
 	PYTHONPATH="." $(PYTHON) -m pytest \
-		-m "not needs_separate_project" \
 		--ignore=tests/runtime/performance \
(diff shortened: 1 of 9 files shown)


SebastianMorawiec (approved):
I dont mind those changes 🤷 it is dev-only marker anyway

PR #96, opened 2023-03-30
docs: guide on workarounds for resource congestion

BenjaminMummery:
[ORQSDK-795]

# The problem

Users need to know how to set CE resources and when they need to override them at the workflow level.

# This PR's solution

Adds a guide giving an overview of resource management and giving troubleshooting advice.

# Checklist

[ORQSDK-795]: https://zapatacomputing.atlassian.net/browse/ORQSDK-795?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


BenjaminMummery:
As always, I've tried to keep the information in the guide minimal and focussed in order to avoid overloading the reader. If there's detail or context you feel is missing, please say so!

docs/guides/ce-resource-management.rst
```diff
@@ -0,0 +1,67 @@
...
+
+A deadlock can be created on Compute Engine if a task attempts to spawn additional processes, notably via the DNQ ``rollouts`` facility. Resources requested in a task definition are bound to the task process, so additional processes can rapidly exhaust the provisioned resources. In these cases, additional resources should be specified in the workflow decorator.
+
+.. code-block::
+    :caption: Example: override workflow resources.
+    @sdk.task(resources=...)                    # task resources requested.
+    def task():
+        config = DQNConfig()
+        ...
+        config.rollouts(num_rollout_workers=2)  # child processes do not have access to
```
jamesclark-Zapata, line 59: This isn't a "child process" in this case. This is using Ray under the hood

docs/guides/ce-resource-management.rst
```diff
@@ -0,0 +1,67 @@
...
+    def my_workflow():
+        ...
+
+In most cases, defining resources in this way will be unnecessary as Compute Engine can infer the overall resource requirements from the aggregated requirements of individual tasks. The primary use-case for this facility is to provision additional resources that aren't covered by the task definitions, such as when tasks spawn additional processes.
+
+
+Troubleshooting Common Resource Issues
+--------------------------------------
+
+A deadlock can be created on Compute Engine if a task attempts to spawn additional processes, notably via the DNQ ``rollouts`` facility. Resources requested in a task definition are bound to the task process, so additional processes can rapidly exhaust the provisioned resources. In these cases, additional resources should be specified in the workflow decorator.
```
jamesclark-Zapata, line 51: Could you mention this is to do with Ray's RLLib?

docs/guides/ce-resource-management.rst
```diff
@@ -0,0 +1,67 @@
...
+.. code-block::
+    :caption: workflow resource request example
+
+    @sdk.workflow(
+        resources=sdk.Resources(cpu="100m", memory="1Gi", disk="10Gi", gpu="1")
+    )
+    def my_workflow():
+        ...
+
+In most cases, defining resources in this way will be unnecessary as Compute Engine can infer the overall resource requirements from the aggregated requirements of individual tasks. The primary use-case for this facility is to provision additional resources that aren't covered by the task definitions, such as when tasks spawn additional processes.
```
jamesclark-Zapata, line 45: Is this implemented yet?

docs/guides/ce-resource-management.rst
```diff
@@ -0,0 +1,67 @@
...
+        resources=sdk.Resources(cpu="100m", memory="1Gi", disk="10Gi", gpu="1")
+    )
+    def my_task():
+        ...
+
+``resources`` expects a ``sdk.Resources()`` object that specifies some or all of:
+* number of cores (``cpu``)
+* amount of RAM (``memory``)
+* disk space (``disk``)
+* number of GPU computing units (``gpu``)
```
jamesclark-Zapata, line 27: This is currently an "on" or "off", with `gpu="1"` or `gpu="0"`

docs/guides/ce-resource-management.rst
```diff
@@ -0,0 +1,67 @@
...
+Setting Workflow Resources
+--------------------------
+
+Resources can also be configured at the workflow definition level using the same syntax as with tasks:
+
+.. code-block::
+    :caption: workflow resource request example
+
+    @sdk.workflow(
+        resources=sdk.Resources(cpu="100m", memory="1Gi", disk="10Gi", gpu="1")
```
jamesclark-Zapata, line 40: Can you explain what the different suffixes mean?
What's "100m" CPUs, how does it relate to a single core. What about "Gi" vs "Mi" vs "G" vs "M" for memory.

docs/guides/ce-resource-management.rst
jamesclark-Zapata, line 45: What about requesting additional nodes?

docs/guides/ce-resource-management.rst
BenjaminMummery: [23c00a1](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/96/commits/23c00a19a968242096a5c1f13b9aa9e12c2d54b7)

docs/guides/ce-resource-management.rst
BenjaminMummery: [9b2fd18](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/96/commits/9b2fd18710b2b04cb2a8e0b085ae691a68d72837)

docs/guides/ce-resource-management.rst
```diff
@@ -21,12 +21,41 @@ Required hardware resources are configured on a per-task basis by setting the ``
...
+    | Metric  | nano  | n      | ~ 2^-29.90 | ~ 1024^-2.990  | 10^-9           |
+    |         | micro | u      | ~ 2^-19.93 | ~ 1024^-1.993  | 10^-6           |
+    |         | milli | m      | ~ 2^-9.966 | ~ 1024^-0.9966 | 10^-3           |
+    |         | kilo  | k      | ~ 2^9.966  | ~ 1024^0.9966  | 10^3            |
+    |         | mega  | M      | ~ 2^19.93  | ~ 1024^1.993   | 10^6            |
+    |         | giga  | G      | ~ 2^29.90  | ~ 1024^2.990   | 10^9            |
+    |         | tera  | T      | ~ 2^39.86  | ~ 1024^3.986   | 10^12           |
+    |         | peta  | P      | ~ 2^49.83  | ~ 1024^4.983   | 10^15           |
+    |         | exa   | E      | ~ 2^59.79  | ~ 1024^5.979   | 10^18           |
+    +---------+-------+--------+------------+----------------+-----------------+
```
jamesclark-Zapata, line 53: The table is kinda complicated. Do we need all three powers?

docs/guides/ce-resource-management.rst
```diff
@@ -21,12 +21,41 @@ Required hardware resources are configured on a per-task basis by setting the ``
...
+
+Amounts of cpu and memory resources can be specified as a plain integer, or as a fixed-point float appended with the string representation of one of the following unit prefixes. These can be SI (metric), or IEC (binary) prefixes.
+
+.. table:: unit prefixes
+    :widths: auto
+
+    +---------+-------+--------+------------+----------------+-----------------+
+    |         | Name  | String | Power of 2 | Power of 1024  | Power of 10     |
+    +=========+=======+========+============+================+=================+
+    | Binary  | kibi  | ki     | 2^10       | 1024^1         | 1.024 x 10^3    |
```
jamesclark-Zapata, line 37: This is a capital "K"?
See https://github.com/kubernetes/apimachinery/blob/v0.26.2/pkg/api/resource/quantity.go#L54

docs/guides/ce-resource-management.rst
```diff
@@ -21,12 +21,41 @@ Required hardware resources are configured on a per-task basis by setting the ``
...
-* disk space (``disk``)
-* number of GPU computing units (``gpu``)
-
-The example above specifies a task that requires 100 milicores (or 0.1 cores), 1 Gb of RAM, 10 Gb of disk space, and access to a GPU.
+* ``cpu``: number of cores.
+* ``memory``: amount of RAM (bytes).
+* ``disk``: disk space (bytes).
+* ``gpu``: whether access to a gpu unit is required (``1`` if a GPU is required, ``0`` otherwise).
+
+Amounts of cpu and memory resources can be specified as a plain integer, or as a fixed-point float appended with the string representation of one of the following unit prefixes. These can be SI (metric), or IEC (binary) prefixes.
```
jamesclark-Zapata, line 29: > Amounts of cpu and memory resources can be specified as a plain integer

They have to be a string, atm.

docs/guides/ce-resource-management.rst
jamesclark-Zapata: This still mentions processes, but it's to do with Ray's resource allocation which is related to actors or remote functions

docs/guides/ce-resource-management.rst
```diff
@@ -0,0 +1,96 @@
...
+    def my_task():
+        ...
+
+``resources`` expects a ``sdk.Resources()`` object that specifies some or all of:
+* ``cpu``: number of cores.
+* ``memory``: amount of RAM (bytes).
+* ``disk``: disk space (bytes).
+* ``gpu``: whether access to a gpu unit is required (``1`` if a GPU is required, ``0`` otherwise).
+
+Amounts of cpu and memory resources can be specified as a plain integer, or as a fixed-point float appended with the string representation of one of the following unit prefixes. These can be SI (metric), or IEC (binary) prefixes. The specified prefix will be used to modify the base unit for the parameter, 'byte' in the case of ``memory`` and ``disk`` requests, 'cores' in the case of ``cpu`` requests. So ``disk="10k"`` will be interpreted as '10 kilobytes', while ``cpu="10k"`` would request 10^7 cores.
```
jamesclark-Zapata, line 29: Suffixes, not prefixes! I made the same mistake the other day 😅

BenjaminMummery: Okay so this is definitely something we need to be careful about the language on. They're prefixes w.r.t. the unit, but suffixes w.r.t. the actual string you write. 'Prefixes' makes sense to me, but as you point out are likely to cause confusion, so we should avoid using the word.
What do you think about:

```suggestion
Amounts of cpu and memory resources are specified by a string comprising a floating point value, and, optionally, a modifier to the base unit ('byte' in the case of ``memory`` and ``disk`` requests, 'cores' in the case of ``cpu`` requests). The modifier can be a SI (metric), or IEC (binary) multiplier as detailed in the table below. So ``disk="10k"`` will be interpreted as '10 kilobytes', while ``cpu="10k"`` would request 10^7 cores.
```

docs/guides/ce-resource-management.rst
BenjaminMummery: "I didn't choose the capitalization" lol
```suggestion
    | Binary  | kibi  | Ki     | 2^10       | 1024^1         | 1.024 x 10^3    |
```

docs/guides/ce-resource-management.rst
BenjaminMummery: Good point, people probably aren't going to want to convert between units all that often. [781f0a8](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/96/commits/781f0a84bf23f3773462a53f80c8b217a74ae054) cuts it down to a single column 'value' and uses 2^n for Binary and 10^n for Metric.

docs/guides/ce-resource-management.rst
jamesclark-Zapata: Ahh yeah you're right! suffix to the "number" but prefix to the "unit".

I think your suggestion is good.

docs/guides/ce-resource-management.rst
```diff
@@ -57,40 +57,46 @@ Convention is to use binary prefixes for memory resource requests (``disk`` and
 .. note:: mixing unit prefixes
     Binary and decimal units can be used interchangeably, however this can occasionally cause confusion, and care must be taken when specifying these parameters. For example, a memory request of ``100m`` specifies not 100 megabytes, but 100 millibytes, or 0.1 bytes.
 
-Setting Workflow Resources
---------------------------
+.. TODO: uncomment and check this section when workflow resource management is implemented (https://zapatacomputing.atlassian.net/browse/ORQSDK-797?atlOrigin=eyJpIjoiNGU1MDU0NjFhNTMxNGUwN2IyZTQzODMxZTVhNjQwM2UiLCJwIjoiaiJ9)
```
jamesclark-Zapata, line 60: I should have mentioned, I implemented this in between making the previous comment and now. That ticket is to do with heterogeneous workflows.

docs/guides/ce-resource-management.rst
BenjaminMummery: [17555c5](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/96/commits/17555c56495c0fb590dd73a2151b5a81803a3b64)

BenjaminMummery: It is now

docs/guides/ce-resource-management.rst
BenjaminMummery: [e150bda](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/96/commits/e150bdad7e314f7da95b1c319fd2d90be78360ac)

docs/guides/ce-resource-management.rst
BenjaminMummery: How about this version: [f9682ac](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/96/commits/f9682ac1f862b88758e76ffcc823976e4540b130) less 'process' more 'actors and remote functions'

docs/guides/ce-resource-management.rst
```diff
@@ -60,18 +60,27 @@ Convention is to use binary prefixes for memory resource requests (``disk`` and
...
     )
     def my_workflow():
         ...
 
-In most cases, defining resources in this way will be unnecessary as Compute Engine can infer the overall resource requirements from the aggregated requirements of individual tasks. The primary use-case for this facility is to provision additional resources that aren't covered by the task definitions, such as when tasks spawn additional actors or remote functions.
+
+.. note:: nodes
+    Note that unlike the other parameters, ``nodes`` must be an integer rather than a string.
+
+In most cases, defining resources in this way will be unnecessary as Compute Engine can infer the overall resource requirements from the aggregated requirements of individual tasks. The primary use-cases for this facility is to provision additional resources that aren't covered by the task definitions, such as when tasks spawn additional actors or remote functions, or to specify a the number of nodes.
```
jamesclark-Zapata, line 83:
```suggestion
Currently, the workflow resource request is only utilised by Compute Engine.
If resources are not provided, Compute Engine will infer the overall resource requirements from the aggregated requirements of individual tasks.
Tweaking the resource request may be required when your tasks spawn additional actors or remote functions to avoid deadlock, see below.
```

jamesclark-Zapata (approved):

PR #112, opened 2023-04-07
feat: print debug information from orq

alexjuda:
# The problem

Recently we've seen a lot of users having issues with fetching data from QE, like logs produced by a workflow. Debugging is difficult because it's not always obvious what HTTP requests are being made.

# This PR's solution

Adds `ORQ_VERBOSE` env variable, used by the `orq` CLI. If set, `orq` will output any log produced by the command internals. This includes HTTP requests made by `requests`, but should also enable more debugging scenarios in the future.

## Example output

```
$ ORQ_VERBOSE=1 orq workflow view -c prod-d
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): prod-d.orquestra.io:443
DEBUG:urllib3.connectionpool:https://prod-d.orquestra.io:443 "GET /v1/workflow?workflowid=orquestra-basic-demo-m0k45-r000 HTTP/1.1" 200 None
DEBUG:urllib3.connectionpool:https://prod-d.orquestra.io:443 "GET /v1/workflow?workflowid=orquestra_basic_demo-8iNWM-r000 HTTP/1.1" 404 19
  File "/Users/alex/Code/zapata/evangelism-workflows/vendor/orquestra-workflow-sdk/src/orquestra/sdk/_base/cli/_dorq/_workflow/_view.py", line 46, in on_cmd_call
    self._on_cmd_call_with_exceptions(*args, **kwargs)
  File "/Users/alex/Code/zapata/evangelism-workflows/vendor/orquestra-workflow-sdk/src/orquestra/sdk/_base/cli/_dorq/_workflow/_view.py", line 56, in _on_cmd_call_with_exceptions
    wf_run = self._wf_run_resolver.resolve_run(wf_run_id, resolved_config)
    ...
```

## Known limitation

We have the `verbose` parameter sprinkled around our codebase (like in `QERuntime`), but it's not accessible by the users since we got rid of `corq`. I'd like to tidy this up separately some time in the future.

# Checklist


tests/cli/dorq/test_cli_logs.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+@pytest.mark.slow()
+class TestConfigureVerbosenessIfNeeded:
+    class TestIntegration:
+        """
+        Python's ``logging`` is notoriously difficult to reason about, so unit tests
+        with mocks don't make much sense. Testing logging configuration with ``pytest``
+        isn't straightforward either because ``pytest`` intercepts log entries. This is
+        why we test the logging config via subprocess & stdout.
```
SebastianMorawiec, line 22: suggestion: sub-process tests are notoriously difficult to debug tho. Could you try to use other means of testing?

src/orquestra/sdk/_base/_env.py
```diff
@@ -47,12 +69,22 @@
...
 RAY_GLOBAL_WF_RUN_ID_ENV = "GLOBAL_WF_RUN_ID"
 """
 Used to set the workflow run ID in a Ray workflow
 """
+
+
+# ------------------------------- utilities ----------------------------------
+
+
+def _is_truthy(env_var_value: t.Optional[str]):
```
SebastianMorawiec, line 81: nitpick: Cant think of any better name of that function, but for me truthy in Python is anything that I can check with if () and is evaluated as true, I.E. list with elements [1].
But maybe the function is so trivial that it is fine 🤔

SebastianMorawiec (approved):
approved with few non blocking comments.

tests/cli/dorq/test_cli_logs.py
alexjuda: Unfortunately I don't know a good way to implement a test like this in a single process. I tried with capsys (https://github.com/zapatacomputing/orquestra-workflow-sdk/blob/ae1cdfbc9e32992281540e74212883b4eacc01da/tests/cli/dorq/test_cli_logs.py#L70-L86) but the stdout/err isn't produced. [Pytest has utils](https://docs.pytest.org/en/7.1.x/how-to/logging.html) for `logging` but they serve a different purpose. Here, I want to ensure we configure the loggers hierarchy correctly, whereas `caplog` allows you to assert that a given function emitted a log record.

src/orquestra/sdk/_base/_env.py
alexjuda: It's about truthy-ness in the context of env variables, as specified by the module name 🙃 I don't have a better idea for this name either, so I'll leave it as is; especially given it's a private function.

PR #115, opened 2023-04-14
fix: Retry getting results from CE after a workflow has succeeded

jamesclark-Zapata:
# The problem

There's a short period of time between a workflow succeeding and the results being available from CE.
The leads to a weird error message:

```
WorkflowRunNotSucceeded: Workflow run `wf_sum_numbers-P1SVm-r000` is in state State.SUCCEEDED
```

# This PR's solution

Allows `CERuntime.get_workflow_run_results_non_blocking` to retry when it raises `WorkflowResultsNotReadyError`.

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_driver/_ce_runtime.py
+++ b/src/orquestra/sdk/_base/_driver/_ce_runtime.py
@@ -210,10 +210,19 @@ class CERuntime(RuntimeInterface):
 
         if len(result_ids) == 0:
             wf_run = self.get_workflow_run_status(workflow_run_id)
-            raise exceptions.WorkflowRunNotSucceeded(
-                f"Workflow run `{workflow_run_id}` is in state {wf_run.status.state}",
-                wf_run.status.state,
-            )
+            if wf_run.status.state == State.SUCCEEDED:
+                raise exceptions.WorkflowResultsNotReadyError(
+                    f"Workflow run `{workflow_run_id}` has succeded, but the results "
+                    "are not ready yet.\n"
+                    "After a workflow completes, there may be a short delay before the "
+                    "results are ready to download. Please try again!"
+                )
+            else:
+                raise exceptions.WorkflowRunNotSucceeded(
+                    f"Workflow run `{workflow_run_id}` is in state "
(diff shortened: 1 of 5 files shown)


src/orquestra/sdk/_base/_driver/_ce_runtime.py
```diff
@@ -208,10 +208,19 @@ def get_workflow_run_outputs_non_blocking(
 
         if len(result_ids) == 0:
             wf_run = self.get_workflow_run_status(workflow_run_id)
-            raise exceptions.WorkflowRunNotSucceeded(
-                f"Workflow run `{workflow_run_id}` is in state {wf_run.status.state}",
-                wf_run.status.state,
-            )
+            if wf_run.status.state == State.SUCCEEDED:
+                raise exceptions.WorkflowResultsNotReadyError(
```
SebastianMorawiec, line 212: how you thought about retrying for a brief sec?
It would be natual for the user to implement their code as:
wf.run()
wf.wait_for_complete()
wf.get_results().

This would give the user the following error. Or maybe we should publish succeeded state only when results are actually ready?

SebastianMorawiec: I think it should with like 1s or so delay so users wont notice, than we can throw an error

src/orquestra/sdk/_base/_retry.py
```diff
@@ -0,0 +1,28 @@
+################################################################################
+# © Copyright 2023 Zapata Computing Inc.
+################################################################################
+import time
+from typing import Optional, Tuple, Type
+
+
+def retry(
```
SebastianMorawiec, line 8: Praise: this looks like a useful function. Good to have it as separate piece of code

SebastianMorawiec (approved):
👍

src/orquestra/sdk/_base/_retry.py
alexjuda: +1. I wanted to suggest something similar when I've seen you discussing retries. Nicely done!

PR #121, opened 2023-04-18
chore: pass for spelling, imports, copyright, style, etc.

BenjaminMummery:
# The problem

We have some typos and formatting inconsistencies in our code.

# This PR's solution

I have a bunch of useful git hooks, so I ran them on everything.

- Fixes every misspelled word I could find
- Removes trailing whitespace
- Sorts the entries in .gitignore
- Removes unused imports
- Marks imports that linters identify as unused but which we actually need (useful for preventing automated import managers from removing them in future)
- Adjusts inconsistent formatting (thank you @jamesclark-Zapata for suggesting Ruff!)
- Ads a copyright marker to files that didn't already have one.

# Checklist

- [X] The PR's title is prefixed with `<feat/fix/chore/internal/docs>[!]:`


src/orquestra/sdk/schema/ir.py
```diff
@@ -82,10 +82,11 @@ class InlineImport(BaseModel):
 
 
 class PackageSpec(BaseModel):
+    # noqa E51
```
jamesclark-Zapata, line 85: what's `E51`?

jamesclark-Zapata (approved):
Only comment: we ignore F401 anyway in the Makefile, does it make sense to remove this ignore?

src/orquestra/sdk/schema/ir.py
BenjaminMummery: It's E501 if you're really lazy!
```suggestion
    # noqa E501
```

BenjaminMummery:
> Only comment: we ignore F401 anyway in the Makefile, does it make sense to remove this ignore?

IMO yes because a) my standard workflow includes a bunch of git hooks that aren't controlled by the makefile and b) it allows us to differentiate between an import that was added by mistake/is no longer needed, and one that we actually want.

alexjuda (approved):
LGTM! Added a couple of non-blocking questions :)

PR #122, opened 2023-04-19
fix: Do not modify arg when building Ray workflow

jamesclark-Zapata:
# The problem

We modified the workflow def's constants when building Ray workflows. Python passes args by ref and this mutated the original workflow def, breaking workflow parsing later.

# This PR's solution

* Copies the requires constants instead of mutating

# Checklist

### DIFF
--- a/src/orquestra/sdk/_ray/_dag.py
+++ b/src/orquestra/sdk/_ray/_dag.py
@@ -383,7 +383,11 @@ def _make_ray_dag(
     client: RayClient, wf: ir.WorkflowDef, wf_run_id: str, project_dir: t.Optional[Path]
 ):
     # We need to modify the dictionary, by adding new key-values.
-    # A shallow copy is OK, as long as we don't modify the _values_ in the dict.
+    # We will need to make a copy to prevent mutating the argument.
+    # Note: this is a *shallow copy* and the values in the dict are not copied but still
+    #       reference the original values. This only works with the assumption that we
+    #       DO NOT MUTATE values in the dictionary.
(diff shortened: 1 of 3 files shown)


src/orquestra/sdk/_ray/_dag.py
```diff
@@ -382,7 +382,9 @@ def _gather_kwargs(
 def _make_ray_dag(
     client: RayClient, wf: ir.WorkflowDef, wf_run_id: str, project_dir: t.Optional[Path]
 ):
-    ray_consts: t.Dict[ir.ConstantNodeId, t.Any] = wf.constant_nodes
+    # We need to modify the dictionary, by adding new key-values.
+    # A shallow copy is OK, as long as we don't modify the _values_ in the dict.
+    ray_consts: t.Dict[ir.ConstantNodeId, t.Any] = wf.constant_nodes.copy()
```
SebastianMorawiec, line 387: Question, non-blocking: would it cost much to make deep-copy just-in-case? Its more error-prone in the future in case we missed whats reference and whats copy. Up to you to decide, not saying it is must-have

SebastianMorawiec (approved):
Non-blocking question added. Up to you to decide

src/orquestra/sdk/_ray/_dag.py
jamesclark-Zapata: After discussing on Slack, we came to the conclusion that shallow copy is OK for now.

Deep copies seemed to be expensive when dealing with workflows with pickled constants (in particular).

I updated the comment to address that we should NOT modify values from the dict.

PR #131, opened 2023-04-25
fix: Custom Image default to None

jamesclark-Zapata:
# The problem

We explicitly set a custom image, even when it was not strictly required.
This may cause conflicts on some new backend changes later.

# This PR's solution

* Switch the default custom image in the Python API to be `None`.
* Update `QE` to pick specific images if the IR custom image is `None`.

# Checklist


alexjuda (approved):
Sooo satisfying! This should've been our implementation from the start!

PR #132, opened 2023-04-25
Orqsdk 801 invalid wf result shape

BenjaminMummery:
# The problem

The behavior of wf_run.get_results() differs between runtimes when the workflow function returns a single, packed artifact.

## Example

```python
@sdk.task
def list_():
    return [1, 2, 3]

@sdk.workflow
def wf1():
    a = list_()
    return a
wf_run = wf1().run(runtime)
wf_run.wait_until_finished()
print("results", wf_run.get_results())
print("artifacts", wf_run.get_artifacts())
print("task outputs", [t.get_outputs() for t in wf_run.get_tasks()])
```

In process:
```
results [1, 2, 3]
artifacts {'invocation-0-task-list-': [1, 2, 3]}
task outputs [[1, 2, 3]]
```

Ray:
```
results ([1, 2, 3],)
artifacts {'invocation-0-task-list-': [1, 2, 3]}
task outputs [[1, 2, 3]]
```

CE:
```
results ([1, 2, 3],)
artifacts {'invocation-0-task-list-': ([1, 2, 3],)}
task outputs [([1, 2, 3],)]
```

QE:
```
results ([1, 2, 3],)
artifacts {'invocation-0-task-list-': [1, 2, 3]}
task outputs [[1, 2, 3]]
```

# This PR's solution

`WorkflowRun.get_results()` checks the expected number of results. Where a single result is expected, it is returned without being enclosed in a sequence. In other words, get_results for a workflow whose result is `"Hello, orquestra!"` now returns `"Hello, orquestra!"` rather than `("Hello, orquestra!",)`. This approach brings the behaviour of `get_results()` into line with vanilla python.

The CLI has previously relied on the length of the output of `get_results` to determine the number of artefacts. This, and any code that uses the same assumption, runs the risk of getting confused if a workflow returns a single artefact that is itself a sequence. Therefore, we add an optional `force_as_sequence` argument to `get_results()` which reproduces the original behaviour. 

`WorkflowRun.get_artifacts()` and `TaskRun.get_outputs()` now check for task outputs that are single-element tuples and unpack these. An alternative solution to this would be performing this check in `ce_runtime.get_available_outputs`, however I wanted to avoid de- and re- serialising the results to perform the check.

# Checklist

- [ ] The [changelog file](CHANGELOG.md) has been updated with a user-readable description of the changes (if the change isn't visible to the user in any way, check the box).
- [X] The PR's title is prefixed with `<feat/fix/chore/internal/docs>[!]:`


jamesclark-Zapata:
@BenjaminMummery we merged the base branch, I can sync with you tomorrow on rebasing it if you'd like!

PR #140, opened 2023-04-28
fix: Set max retries in Ray

jamesclark-Zapata:
# The problem

When a Ray worker crashes, it will try to run the task again. This causes issues when using things that assume a single execution of a task.

# This PR's solution

* Set `max_retries` to 0.

# Checklist

### DIFF
--- a/tests/runtime/ray/test_client.py
+++ b/tests/runtime/ray/test_client.py
@@ -26,6 +26,7 @@ class TestClient:
                 "metadata": {},
                 "catch_exceptions": False,
                 "runtime_env": None,
+                "max_retries": 0,
             }
 
         def test_required_args(self, client: RayClient, remote_fn, required_kwargs):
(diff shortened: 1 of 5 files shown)


tests/runtime/ray/test_client.py
```diff
@@ -22,6 +22,7 @@ def required_kwargs(self):
                 "metadata": {},
                 "catch_exceptions": False,
                 "runtime_env": None,
+                "max_retries": 0,
```
alexjuda, line 25: **suggestion**: please add an inline comment stating we need this option to workaround issues with duplicated run IDs when a Ray worker crashes. I know it's in the PR & ticket descriptions but a future dev might easily overlook this.

alexjuda (approved):
LGTM! Added a suggestion.

tests/runtime/ray/test_client.py
jamesclark-Zapata: There's a comment for w/ with invocation `max_retries`, I added a comment here to reference that.

PR #142, opened 2023-05-03
[OP-39] orquestra-sdk-base images for CUDA and non-CUDA workflows

emre-aydin:
This PR creates the Dockerfiles for orquestra-sdk-base images along with its integration to the CI/CD Docker image build infrastructure. 

Each new tag will trigger creation of new images for these builds. There's also an option to specify the SDK version manually to build development images.

Building new images with each tag might be problemmatic if tags are created before publishing Python modules to PyPI. Let me know if this is the case so that I can remove those triggers from the build and find another way to tag the images.


cuda.Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+apt install --yes cmake
+EOF
+
+# install qsimcirq and orquestra SDK
+RUN <<EOF
+git clone https://github.com/quantumlib/qsim.git
+cd qsim
+make clean
+make
+python -m pip install -e .
```
roryschadler, line 47: I was talking with @brian-dellabetta about this, I wonder if we really need to do this from scratch? It adds a significant amount of time to the build process for the notebooks. If the pre-built wheels don't exist, or aren't adequate, my proposal is to build it once, and then upload our "custom" wheel to our private Nexus PyPI repository.

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+apt install --yes --no-install-recommends git
+python -m pip install -U pip wheel
+python -m pip install orquestra-sdk[ray]==${SDK_VERSION}
+EOF
+
+RUN useradd -ms /bin/bash -d /home/orquestra orquestra --uid 1000 --gid 100
+USER 1000
+WORKDIR /home/orquestra
+
+ENV RAY_STORAGE=/tmp
```
roryschadler, line 17: Are these workflow-driver specific env vars?

cuda.Dockerfile
emre-aydin: It sounds like this could be improved later on. Hence I suggest to not do it here (to avoid scope creep) and improve it after this PR gets merged.

Dockerfile
emre-aydin: AFAIK they're not. @SebastianMorawiec can confirm.

Dockerfile
```diff
@@ -0,0 +1,23 @@
+# syntax=docker/dockerfile:1.5
+FROM python:3.9-slim-bullseye
+ARG SDK_VERSION
+
+RUN <<EOF
+apt update --yes
+apt upgrade --yes
+apt install --yes --no-install-recommends git
+python -m pip install -U pip wheel
+python -m pip install orquestra-sdk[ray]==${SDK_VERSION}
```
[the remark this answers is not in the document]
emre-aydin: No it won't. I have considered COPYing the repo in but wasn't sure it would be better because we will need to (at least in the beginning) create images for past SDK versions and having the SDK version as a parameter makes it more flexible. Depending on the SDK release process, we can convert to COPYing the repo. WDYT @zapatacomputing/sdk-team ?

Dockerfile
SebastianMorawiec: ORQ_RAY_DOWNLOAD_GIT_IMPORTS is used in SDK overall, should be only set when running on remote.
ORQ_RAY_SET_TASK_RESOURCES was removed by me. Can be removed here as well I believe

Dockerfile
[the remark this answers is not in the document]
alexjuda: >  I have considered COPYing the repo in but wasn't sure it would be better because we will need to (at least in the beginning) create images for past SDK versions and having the SDK version as a parameter makes it more flexible.

That `*` in the `.dockerignore` got me wondering too but it's quite clever! 

How about something like this?
```suggestion
python -m pip install "${SDK_REQUIREMENT}"
```

It would allow us to install published versions from PyPI with `SDK_REQUIREMENT='orquestra-sdk[ray]==0.47.0'` as well as tagged, but not published yet versions like `SDK_REQUIREMENT='orquestra-sdk[ray] @ git+https://github.com/zapatacomputing/orquestra-workflow-sdk.git@v0.47.0'`

Dockerfile
```diff
@@ -0,0 +1,23 @@
+# syntax=docker/dockerfile:1.5
```
alexjuda, line 1: **nitpick**: could you state the expected usage of images built with these dockerfiles? Something like

```suggestion
# syntax=docker/dockerfile:1.5
# Base image for running Orquestra tasks with custom environment.
```

alexjuda, line 1: **suggestion**: WDYT about moving the docker-related stuff to a dedicated subdir `docker/`? This layout comes handy when there's more docker-related configuration required, and changing it in the future might be cumbersome.

cuda.Dockerfile
```diff
@@ -0,0 +1,61 @@
+# syntax=docker/dockerfile:1.5
```
alexjuda, line 1: Plz see comments about dir layout and dockerfile documentation :)

.github/workflows/publish-docker-image.yml
```diff
@@ -67,6 +71,6 @@ jobs:
           # as on the command line `--build-arg`, one to a line:
           # docker_build_args: BUILD_ARG_ONE=value1\nBUILD_ARG_TWO=value2
           # DUE TO JSON LIMITATIONS, NEWLINES MUST BE EXPLICITLY ADDED AS \n
-          docker_build_args: 'SDK_VERSION=${{ inputs.sdk_version }}'
+          docker_build_args: SDK_VERSION=${{ github.ref_type == 'tag' && github.ref_name || inputs.sdk_version }}
```
alexjuda, line 74: @emre-aydin 
> Building new images with each tag might be problemmatic if tags are created before publishing Python modules to PyPI.

You're right that we create tags before we publish wheels to PyPI. For more details see steps 3 and 9 in the [SDK release process](https://zapatacomputing.atlassian.net/wiki/spaces/ORQSRUN/pages/461045791/SDK+release+process).

> Let me know if this is the case so that I can remove those triggers from the build and find another way to tag the images.

I think we could still make it work by installing from git source with something like 

```
SDK_REQUIREMENT="orquestra-sdk[ray] @ git+https://github.com/zapatacomputing/orquestra-workflow-sdk.git@{{github.ref_name}}"
```

Unfortunately I don't know GitHub's syntax enough to suggest a full ternary operator + string interpolation expression.

Alternatively, we can trigger it separately when the wheels are available on PyPI (step 9 in the [release process](https://zapatacomputing.atlassian.net/wiki/spaces/ORQSRUN/pages/461045791/SDK+release+process)).

.dockerignore
```diff
@@ -0,0 +1,13 @@
+# Ignore everything
```
alexjuda, line 1:
```suggestion
# Ignore everything. We install the image contents separately from the repo the
# Dockerfile is contained in. This lets us use a single Dockerfile with
# different versions of Orquestra Workflow SDK.
```

alexjuda (changes requested):

cuda.Dockerfile
alexjuda: Another alternative to consider in the future is split the `qsim` and `orquestra-sdk` installations to separate Docker layers + pin `qsim`'s version so that its installation would be cached.

alexjuda, line 47: Minor suggestion:
```suggestion
python -m pip install .
```
The editable mode sometimes does funky things. I think it's better to avoid it when we're not suiting up `qsim` for development of `qsim` itself.

Dockerfile
emre-aydin: Sure, will remove.

Dockerfile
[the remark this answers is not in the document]
emre-aydin: Exactly what I'm looking for. IIUC this will also let us use development versions of SDK in CE and we'll be able to mark [the task we have for this on SDK backlog](https://zapatacomputing.atlassian.net/browse/ORQSDK-833) as completed, right?

Dockerfile
emre-aydin: Makes sense.

.github/workflows/publish-docker-image.yml
emre-aydin: I would prefer having things as flexible as possible in the beginning so that we can learn and adapt the processes as we go. Let me remove the tag trigger and manually trigger the build as part of the release process in the beginning.

cuda.Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+# Mounted gpu has cuda v11.5
+FROM nvcr.io/nvidia/cuquantum-appliance:22.03-cirq
```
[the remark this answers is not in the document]
roryschadler: I believe this was borrowed from `orquestra-notebooks`, which was in turn borrowed from `orquestra-nvidia-image`. I've added a ticket to review: https://zapatacomputing.atlassian.net/browse/OP-47

.github/workflows/publish-cuda-docker-image.yml
```diff
@@ -1,14 +1,16 @@
...
-      sdk_version:
+      sdk_requirement:
         type: string
-        description: Orquestra SDK version to build the image with
+        description: |
+          Orquestra SDK requirement to build the image with. 
+          Example of using a published version of the SDK:
+          orquestra-sdk[ray]==0.47.0 
+          Example of using an unpublished version:
+          orquestra-sdk[ray] @ git+https://github.com/zapatacomputing/orquestra-workflow-sdk.git@v0.47.0
```
alexjuda, line 13: **praise**: that's a very helpful description, thanks!

Dockerfile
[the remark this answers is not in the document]
alexjuda: > IIUC this will also let us use development versions of SDK in CE and we'll be able to mark [the task we have for this on SDK backlog](https://zapatacomputing.atlassian.net/browse/ORQSDK-833) as completed, right?

```
             ┌─2──────┐
         ┌──►│Task Run│
         │   └────────┘
  ┌─1─┐  │   ┌─3──────┐
  │WDR├──┴──►│Task Run│
  └───┘      └────────┘
```

Dockerfiles from this PR serve the purpose of running tasks with custom images. CE needs a WDR image (1 on the diagram). Is it gonna be the same as the image used for running tasks (2 and 3 on the diagram)? I think think the answer is "no", but wanna make sure.

If I'm right, unfortunately https://zapatacomputing.atlassian.net/browse/ORQSDK-833 is a separate thing because it's about dev versions of WDR images.

alexjuda (approved):
LGTM!

emre-aydin:
> Dockerfiles from this PR serve the purpose of running tasks with custom images.

@alexjuda - not only that. My plan is to use these images as the default images for running workflows. The workflow driver ray image will only run on the head node. Worker nodes will run the images in this PR (or images extended from these images). You're right though, we also need WDR versions of this to be able to test effects of changes to SDK on workflow-driver-ray.

PR #144, opened 2023-05-04
internal: allow Ray 2.4.0

alexjuda:
# The problem

Ray 2.4.0 was released. We had a strict requirement on 2.3.0. It wasn't possible to use 2.4.0 unless you depend on pip weirdness.

# This PR's solution

Loosens the requirement on Ray so `orquestra-sdk` can be used with both Ray 2.3.0 and 2.4.0. If a user project doesn't work with Ray 2.4.0, it should be added as a project requirement instead of our requirement.

This might also help with hanging Windows tests.

# Checklist


alexjuda:
We're gonna use https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/137 to avoid problems with pickling.

PR #145, opened 2023-05-04
internal: Tagging Docker images via optional parameter [OP-39]

emre-aydin:
# The problem

It's not possible to tag Docker images based on the SDK version with the current version of the GitHub Actions workflow that we have for publishing Docker images.

# This PR's solution

Add an optional parameter to the GitHub Actions workflow that publishes Docker images to optionally tag the published Docker image. This way, when we want to publish a Docker image and also tag it with the version of SDK that it uses, we'll be able to do so.

# Checklist

### DIFF
--- a/.github/workflows/publish-docker-image.yml
+++ b/.github/workflows/publish-docker-image.yml
@@ -11,6 +11,11 @@ on:
           orquestra-sdk[ray]==0.47.0 
           Example of using an unpublished version:
           orquestra-sdk[ray] @ git+https://github.com/zapatacomputing/orquestra-workflow-sdk.git@v0.47.0
+      docker_tag:
+        type: string
+        description: |
+          Optional parameter to tag the Docker image. If provided,
+          the published image will be tagged with the given tag.
 jobs:
   trigger-build-and-push:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-docker-image.yml
```diff
@@ -59,7 +64,7 @@ jobs:
           # See https://github.com/docker/metadata-action#tags-input for the
           # format this should take.
           # DUE TO JSON LIMITATIONS, NEWLINES MUST BE EXPLICITLY ADDED AS \n
-          additional_image_tags: ''
+          additional_image_tags: 'type=raw,value=${{ inputs.docker_tag }}'
```
alexjuda, line 67: **non-blocking,question**: Docker image names come with "repo" and "tag" parts, but sometimes the nomenclature is ambiguous. I can see `target_docker_repository` is specified separately above. Just to make sure, what's gonna be the end "full image name" when I run this workflow with `docker_tag` set to `foo-123`?
1. `hub.stage.nexus.orquestra.io/zapatacomputing/orquestra-sdk-base:foo-123`
2. `foo-123`

alexjuda (approved):
The diffs LGTM, but I haven't run the workflow myself. Added one non-blocking question.

.github/workflows/publish-docker-image.yml
emre-aydin: It will be the first one. Added more clarification to the parameter description.

alexjuda (approved):

PR #146, opened 2023-05-05
Internal, Fix: rewrite tests to use sockets instead of FS

SebastianMorawiec:
# The problem

Some of our tests were stuck for some reason.

# This PR's solution

 Rewriting IPC to sockets to see if this makes them more reliable

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_testing/_ipc.py
+++ b/src/orquestra/sdk/_base/_testing/_ipc.py
@@ -1,7 +1,7 @@
-import socket
 import select
-import typing as t
+import socket
 import time
+import typing as t
 
 TRIGGER_MSG = "t".encode()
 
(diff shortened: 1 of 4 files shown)


src/orquestra/sdk/_base/_testing/_ipc.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def close(self):
+        assert self._socket is not None
+
+        self._socket.close()
+
+
+class TriggerClient:
+    def __init__(self, port: int):
+        self._port = port
```
alexjuda, line 42: **nitpick**: looks like `TriggerClient` is used only inside Ray worker processes during tests so `pytest` doesn't register it in coverage stats. Can you mark it as `nocover`?

alexjuda (approved):
Good work! Added nitpicks

PR #147, opened 2023-05-05
fix: `orq submit` with dirty repo

alexjuda:
# The problem

`orq wf submit` would fail with type error if the git repo was dirty.

# This PR's solution

* Passes the missing arguments in the CLI action.
* Makes the tests more robust so we catch bugs like this in the future:
    * Replaces `Mock()` with `create_autospec`.
    * Uses `mock.method.side_effect = ...` instead of `mock.method = ...`. This allows to make use of autospec's signature checks.

I'm not editing the changelog because `orq wf submit` with workspaces support wasn't released yet so users won't notice this bug.

# Checklist


SebastianMorawiec (approved):
Good catch. We definitely should opt for some higher-level testing of the CLI tbh.

PR #148, opened 2023-05-05
feat: support for custom images on CE

alexjuda:
# The problem

Sometimes users need to provide custom docker images with external deps to run their Orquestra tasks. It was possible with QE but not with CE yet.

# This PR's solution

Makes use of Ray's custom resources to embed the custom image name a user specified either:
* in the task def: `@sdk.task(custom_image="foo/bar:abc")`
* in the task invocation inside the workflow: `future = my_task(...).with_invocation_meta(custom_image="foo/bar:abc")`

## Known limitation

If a task in the workflow IR specifies a custom image and you attempt to run the workflow on a local `orq up`-style Ray cluster, Ray will never schedule task for execution. That's the reason why I needed to update the `Test3rdParty` integration test—it used an IR generated with an old SDK version from a time when we embedded `zapatacomputing/orquestra-default:v0.0.0` to every task def. 

@SebastianMorawiec and I talked a bit to figure out a sensible solution. We couldn't find any obvious one, so I added a separate ticket to discuss our options: https://zapatacomputing.atlassian.net/browse/ORQSDK-838

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,12 @@
 * Removed deprecated save() method from RuntimeConfig class
 * Removed is_saved() method and "name" setter from RuntimeConfig class
 * WorkflowRun.get_results() returns values consistent with vanilla python - single results are returned as-as, multiple results are returned as a tuple.
+* If you set a custom task image (`@sdk.task(custom_image=...)`), the local Ray runtime won't execute the task unless the local Ray cluster was started with custom resources.
 
 🔥 *Features*
 * New API functions: list_workspaces() and list_projects(). Usable only on CE runtime
 * Setting workflow_id and project_id is now available using "orq wf submit" command
+* Support for running tasks in Docker containers with custom images on Compute Engine.
 
 👩‍🔬 *Experimental*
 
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -7,10 +7,12 @@
 * Removed deprecated save() method from RuntimeConfig class
 * Removed is_saved() method and "name" setter from RuntimeConfig class
 * WorkflowRun.get_results() returns values consistent with vanilla python - single results are returned as-as, multiple results are returned as a tuple.
+* If you set a custom task image (`@sdk.task(custom_image=...)`), the local Ray runtime won't execute the task unless the local Ray cluster was started with custom resources.
```
SebastianMorawiec, line 10: Suggestion: Please do add a jira ticket to do something with that, either warn a user or change the behaviour to ignore custom image locally entirely.

SebastianMorawiec (approved):
Looks good, added 1 suggestion

CHANGELOG.md
alexjuda: The follow up ticket was mentioned in the PR description: https://zapatacomputing.atlassian.net/browse/ORQSDK-838

I didn't add it to the changelog because I don't think it fits here.

PR #150, opened 2023-05-09
feat: add sdk.get_backend_ids() function.

BenjaminMummery:
[ORQSDK-823]

# The problem

Users who want to track their metrics in MLflow need to somehow identify MLflow experiments and MLflow runs. “Tagging” MLflow stuff with WorkflowRunID/TaskInvocationID could be used to correlate MLflow entities with Orquestra ones. At the moment there’s no good way to get this contextual information from within the user-written task code.

# This PR's solution

Adds a new function `sdk.get_backend_ids()` to the API. This function does the following:
- Determines the current backend - Argo, Ray, or in-process
    - Argo backends are identified by the presence of "ARGO_NODE_ID" in the environment variables
    - Ray backends are identified by the presence of "function='_ray_remote'" in the call stack (it's there for local ray instances as well)
    - In process backends are identified by the absence of the identifiers for ray and argo.
- Determines the current WFRunID, TaskRunID, and TaskInvID
    - For Argo, these are determined from the componenets of the Node ID
    - For Ray, we use _dag.get_current_ids()
    - For In process, we use the IDS global variable, see below
- returns a tuple of these values.

In order to access the ids from within a task, we add a global variable IDS to the in-process runtime. This is set by a context manager for each task execution. `sdk.get_backend_ids()` then imports this at execution time, and returns its value.

## Example usage:

```python
import sdk

@sdk.task
def t():
    wf_run_id, task_inv_id, task_run_id =  sdk.get_backend_ids()
    ...
```

# Checklist

[ORQSDK-823]: https://zapatacomputing.atlassian.net/browse/ORQSDK-823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/orquestra/sdk/_base/_api/_task_run.py
+++ b/src/orquestra/sdk/_base/_api/_task_run.py
@@ -341,7 +341,13 @@ def get_backend_ids() -> (
     ]
 ):
     """
-    Get the workflow run, task invocation, and task run IDs for the current task.
+    Get the workflow run, task invocation, and task run IDs related to current execution context. 
+    
+    Workflow run ID is a globally unique identifier generated whenever a workflow is submitted for running. Single workflow definition can be run multiple times resulting in multiple workflow run IDs. Analog of PID for a standard program.
+    
+    Task invocation ID is related to using a task in your workflow definition, analogous to a function invocation in a standard program. Scoped to a workflow definition. Isn't globally unique. If you run the same workflow definition multiple times, you'll end up with the same task invocation IDs across runs.
+    
+    Task run ID is a globally unique identifier of executing an invocation exactly once.
 
     This function is intended to be used within the task code in the following way:
     ```
--- a/src/orquestra/sdk/_base/_api/_task_run.py
+++ b/src/orquestra/sdk/_base/_api/_task_run.py
@@ -262,7 +262,9 @@ class TaskRun:
         return set(parents)
 
 
-def _get_argo_backend_ids() -> t.Tuple[WorkflowRunId, TaskInvocationId, TaskRunId]:
+def _get_argo_backend_ids() -> (
+    t.Tuple[WorkflowRunId, t.Optional[TaskInvocationId], t.Optional[TaskRunId]]
+):
     """
     Get the workflow run, task invocation, and task run IDs from Argo.
 
@@ -291,7 +293,9 @@ def _get_argo_backend_ids() -> t.Tuple[WorkflowRunId, TaskInvocationId, TaskRunI
     return wf_run_id, task_inv_id, task_run_id
 
 
-def _get_ray_backend_ids() -> t.Tuple[WorkflowRunId, TaskInvocationId, TaskRunId]:
+def _get_ray_backend_ids() -> (
+    t.Tuple[WorkflowRunId, t.Optional[TaskInvocationId], t.Optional[TaskRunId]]
+):
     """
     Get the workflow run, task invocation, and task run IDs from Ray.
 
@@ -311,7 +315,7 @@ def _get_ray_backend_ids() -> t.Tuple[WorkflowRunId, TaskInvocationId, TaskRunId
 
 
 def _get_in_process_backend_ids() -> (
-    t.Tuple[WorkflowRunId, TaskInvocationId, TaskRunId]
+    t.Tuple[WorkflowRunId, t.Optional[TaskInvocationId], t.Optional[TaskRunId]]
 ):
     """
     Get the workflow run, task invocation, and task run IDs from the In-process runtime.
@@ -335,9 +339,7 @@ def _get_in_process_backend_ids() -> (
 
 
 def current_run_ids() -> (
-    t.Tuple[
-        t.Optional[WorkflowRunId], t.Optional[TaskInvocationId], t.Optional[TaskRunId]
-    ]
+    t.Tuple[WorkflowRunId, t.Optional[TaskInvocationId], t.Optional[TaskRunId]]
 ):
     """
     Get the workflow run, task invocation, and task run IDs related to current
--- a/src/orquestra/sdk/_base/_api/_task_run.py
+++ b/src/orquestra/sdk/_base/_api/_task_run.py
@@ -325,12 +325,18 @@ def current_run_ids() -> (
     ]
 ):
     """
-    Get the workflow run, task invocation, and task run IDs related to current execution context. 
-    
-    Workflow run ID is a globally unique identifier generated whenever a workflow is submitted for running. Single workflow definition can be run multiple times resulting in multiple workflow run IDs. Analog of PID for a standard program.
-    
-    Task invocation ID is related to using a task in your workflow definition, analogous to a function invocation in a standard program. Scoped to a workflow definition. Isn't globally unique. If you run the same workflow definition multiple times, you'll end up with the same task invocation IDs across runs.
-    
+    Get the workflow run, task invocation, and task run IDs related to current
+    execution context.
+
+    Workflow run ID is a globally unique identifier generated whenever a workflow is
+    submitted for running. Single workflow definition can be run multiple times
+    resulting in multiple workflow run IDs. Analog of PID for a standard program.
+
+    Task invocation ID is related to using a task in your workflow definition,
+    analogous to a function invocation in a standard program. Scoped to a workflow
+    definition. Isn't globally unique. If you run the same workflow definition multiple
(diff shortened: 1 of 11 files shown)


BenjaminMummery:
Three questions worth discussion:
- Is _task_run the correct place for this or does it want to be its own thing? I put in _task_run because it makes sense in the context of a specific run, but it doesn't really relate to anything else in there.
- is `get_backend_ids` sufficiently clear and self explanatory for a user-facing function?
- Is there a good way to test the deferred import of the IDS global short of actually running a workflow?

src/orquestra/sdk/_base/_api/_task_run.py
```diff
@@ -257,3 +260,111 @@ def get_parents(self) -> t.Set["TaskRun"]:
...
+
+def get_backend_ids() -> (
+    t.Tuple[
+        t.Optional[WorkflowRunId], t.Optional[TaskInvocationId], t.Optional[TaskRunId]
+    ]
+):
+    """
+    Get the workflow run, task invocation, and task run IDs for the current task.
+
+    This method is intended to be used within the task code in the following way:
```
alexjuda, line 346: **non-blocking,nitpick**: it's not a method
```suggestion
    This function is intended to be used within the task code in the following way:
```

src/orquestra/sdk/_base/_in_process_runtime.py
```diff
@@ -20,6 +33,30 @@
 ArtifactValue = t.Any
 TaskOutputs = t.Tuple[ArtifactValue, ...]
 
+IDS: t.Optional[t.Tuple[WorkflowRunId, ir.TaskInvocationId, TaskRunId]] = None
```
alexjuda, line 36: **non-blocking,nitpick**: I think using globals here is indeed the correct, pragmatic approach. 

By convention, `ALL_CAPS` variables are constants. Meanwhile, the run identifiers change all the time. Would this be a better name?
```suggestion
current_run_ids: t.Optional[t.Tuple[WorkflowRunId, ir.TaskInvocationId, TaskRunId]] = None
```

alexjuda (changes requested):

src/orquestra/sdk/_base/_in_process_runtime.py
BenjaminMummery: Good idea, [34f7c7b](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/150/commits/34f7c7b4b799dc2658cc0691577a61c38d606dd3)

src/orquestra/sdk/_base/_api/_task_run.py
```diff
@@ -257,3 +260,111 @@ def get_parents(self) -> t.Set["TaskRun"]:
...
+    return "ARGO_NODE_ID" in os.environ
+
+
+def _is_ray_backend():
+    """
+    Ray backend test.
+
+    Checks the call stack for `_ray_remote`.
+    """
+    return "function='_ray_remote'" in str(inspect.stack())
```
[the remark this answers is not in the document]
BenjaminMummery: Much nicer, [cac2cb1](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/150/commits/cac2cb1ce264590fcb5ebac3ba3d887d6f06c7c1)

src/orquestra/sdk/_base/_log_adapter.py
```diff
@@ -164,12 +124,11 @@ def workflow_logger() -> logging.LoggerAdapter:
...
     task_run_id: t.Optional[TaskRunId]
-    if is_argo_backend():
-        # Workflow is running in the Orquestra QE environment
-        wf_run_id, task_inv_id, task_run_id = get_argo_backend_ids()
-    else:
-        # We assume the workflow is running on Ray
-        wf_run_id, task_inv_id, task_run_id = get_ray_backend_ids()
+    try:
+        wf_run_id, task_inv_id, task_run_id = current_run_ids()
+    except ModuleNotFoundError:
```
alexjuda, line 129: **question**: Is is still part of the `current_run_ids()` contract?

src/orquestra/sdk/exceptions.py
```diff
@@ -204,6 +204,12 @@ class WorkflowResultsNotReadyError(NotFoundError):
     """
 
 
+class WorkflowRunIDNotFoundError(NotFoundError):
+    """Raised when we can't recover the ID for this Workflow Run."""
+
+    # Note that this is different to the WorkflowRunNotFoundError.
```
alexjuda, line 210: **praise**: good comment!

src/orquestra/sdk/exceptions.py
```diff
@@ -204,6 +204,12 @@ class WorkflowResultsNotReadyError(NotFoundError):
     """
 
 
+class WorkflowRunIDNotFoundError(NotFoundError):
+    """Raised when we can't recover the ID for this Workflow Run."""
+
```
alexjuda, line 209: **non-blocking,suggestion**: WDYT about keeping the `wf_run_id` as an attribute of this error? It's a pattern I'd like to adopt whenever adding exceptions because it allows easier pretty-printing by callers like the CLI.

BenjaminMummery: Given that its raised when we can't recover the workflow run ID, I'm not sure what utility that would serve?

src/orquestra/sdk/_base/_api/_task_run.py
```diff
@@ -257,3 +260,128 @@ def get_parents(self) -> t.Set["TaskRun"]:
...
+    wf_run_id = "-".join(node_id.split("-")[:-2])
+    task_run_id = node_id
+
+    argo_template = json.loads(os.environ["ARGO_TEMPLATE"])
+    # Looks like the template name on Argo matches our task invocation ID. Not sure how
+    # good this assumption is.
+    task_inv_id = argo_template["name"]
+
+    if wf_run_id is None:
+        raise WorkflowRunIDNotFoundError("Could not recover Workflow Run ID")
```
alexjuda, line 291: **issue**: is this check ever going to pass? Even when `node_id.split("-")[:-2]` is `[]`, `"-".join([])` returns `""`

src/orquestra/sdk/exceptions.py
alexjuda: Oooooooh, you're right. We can't assign `self.wf_run_id` because we can't infer it, and it's exactly what's this error is about. Nevermind!

src/orquestra/sdk/_base/_api/_task_run.py
BenjaminMummery: Good point,  changed to check for empty strings insteada3ca472

src/orquestra/sdk/exceptions.py
BenjaminMummery: Thanks, you just reminded me I wanted to explain it further! Changed in [b233180](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/150/commits/b2331809317a9d62c4601a6434a39c16a963711c) to read:
```
# Note that this is different to the WorkflowRunNotFoundError.
# We have an ID but can't find the workflow: WorkflowRunNotFoundError
# We have a workflow but can't recover the ID: WorkflowRunIDNotFoundError
```

src/orquestra/sdk/_base/_api/_task_run.py
```diff
@@ -366,6 +366,10 @@ def t():
 
     Returns:
         WorkflowRunId, TaskInvocationId, TaskRunId
+
+    Raises:
+        ModuleNotFoundError: raised from _get_ray_backend_ids when ray is not installed.
```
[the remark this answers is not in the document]
BenjaminMummery: The workaround for the weird _exec_ctx.global_context behaviour I found is that if _get_in_process_backend_ids() doesn't work, we try _get_ray_backend_ids() instead. This means that there are situations where we try to import ray despite the global_contex telling us its DIRECT, so I'd err on the side of caution at least until I've got to the bottom of what's happening there! 

How's this for a workaround: 
- Remove ModuleNotFoundError from the docstrings
- Catch the ModuleNotFoundError in `_task_run._get_ray_backend_ids` and raise a WorkflowRunIDNotFoundError from it.

WDYT?

alexjuda (approved):
LGTM! Thanks for getting into the root of the issue with me :)

PR #152, opened 2023-05-11
internal: Fix CUDA image publishing workflow failure when no Docker tag is given

emre-aydin:
# The problem

When no Docker tag is given as an input parameter, the GitHub Actions workflow that builds CUDA images fails. However, it needs to be an optional input parameter and the image should be tagged with prefix `dev-build` instead.

# This PR's solution

Modify GitHub Actions workflow code to add the additional tags only when they're provided.

# Checklist

### DIFF
--- a/.github/workflows/publish-cuda-docker-image.yml
+++ b/.github/workflows/publish-cuda-docker-image.yml
@@ -15,9 +15,9 @@ on:
         type: string
         description: |
           Optional parameter to tag the Docker image. If provided,
-          the published image will be tagged with the given tag.
+          the published image will be tagged with the given tag plus -cuda.
           For e.g. if foo-123 is given, the final image will be named
-          hub.stage.nexus.orquestra.io/zapatacomputing/orquestra-sdk-base:foo-123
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-cuda-docker-image.yml
```diff
@@ -66,7 +66,7 @@ jobs:
           # See https://github.com/docker/metadata-action#tags-input for the
           # format this should take.
           # DUE TO JSON LIMITATIONS, NEWLINES MUST BE EXPLICITLY ADDED AS \n
-          additional_image_tags: 'type=raw,value=${{ inputs.docker_tag }}'
+          additional_image_tags: ${{ inputs.docker_tag && format('type=raw,value={0}', inputs.docker_tag) || ''}}
```
alexjuda, line 69: Similar question like in the other repo. Shouldn't this append `-cuda`?
```suggestion
          additional_image_tags: ${{ inputs.docker_tag && format('type=raw,value={0}', inputs.docker_tag)-cuda || ''}}
```

emre-aydin: Same as in the other comment.

alexjuda (approved):

PR #159, opened 2023-05-18
internal: Add a makefile for building the sdk images, tweak for building on M1

roryschadler:
# The problem

Studio development requires building images locally, which is _much_ faster if it's an ARM-native image. Also, Makefiles remove the need to remember `docker build` commands.

Thirdly, workflow dispatches have a limit of 100 chars on the `event_type` field 🥲 

# This PR's solution

Added a Makefile, and installed a few more dependencies in the image build, conditional on the `TARGETARCH` [automatic `ARG`](https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope). Tweaked the `event_type` field in the CICD-actions dispatches.

# Checklist

_Check that this PR satisfies the following items:_

- [x] Tests have been added for new features/changed behavior (if no new features have been added, check the box).
- [x] The [changelog file](CHANGELOG.md) has been updated with a user-readable description of the changes (if the change isn't visible to the user in any way, check the box).
- [x] The PR's title is prefixed with `<feat/fix/chore/internal/docs>[!]:`
- [x] The PR is linked to a JIRA ticket (if there's no suitable ticket, check the box).

PR #160, opened 2023-05-18
fix: Reduce CPU base image size

jamesclark-Zapata:
Based on #159

# The problem

The image layer that does the dependency installation keeps some unwanted directories behind.

# This PR's solution

* Installs from PyPI without cache (~150 MB reduction in image size)
* Removes apt lists after use (~20 MB reduction)
* Uses `apt-get` instead of `apt`, `apt` does not have a stable interface.

Image size built on my laptop:
Before: 798.34 MB
After: 637.19 MB

# Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -11,16 +11,18 @@ ARG TARGETARCH
 # (development using an M1 is fun :') )
 RUN <<EOF
 set -ex
-apt update --yes
-apt upgrade --yes
-apt install --yes --no-install-recommends git openssh-client
+apt-get update --yes
+apt-get upgrade --yes
+apt-get install --yes --no-install-recommends git openssh-client
 
 if [ "$TARGETARCH" = "arm64" ]; then
-  apt install --yes --no-install-recommends gcc libc-dev
+  apt-get install --yes --no-install-recommends gcc libc-dev
 fi
 
-python -m pip install -U pip wheel
-python -m pip install "${SDK_REQUIREMENT}"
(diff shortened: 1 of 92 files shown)


docker/Dockerfile
```diff
@@ -4,12 +4,25 @@
...
+set -ex
+apt-get update --yes
+apt-get upgrade --yes
+apt-get install --yes --no-install-recommends git openssh-client
+
+if [ "$TARGETARCH" = "arm64" ]; then
+  apt-get install --yes --no-install-recommends gcc libc-dev
+fi
+
+rm -rf /var/lib/apt/lists/*
```
roryschadler, line 22: **suggestion**: A few more cleanup commands (Doubt they'll save much more, but maybe just for completeness)
```suggestion
apt-get autoremove --yes
apt-get clean
rm -rf /var/lib/apt/lists/*
```

jamesclark-Zapata: > Official Debian and Ubuntu images [automatically run apt-get clean](https://github.com/moby/moby/blob/03e2923e42446dbb830c654d0eec323a0b4ef02a/contrib/mkimage/debootstrap#L82-L105), so explicit invocation is not required.

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get

I don't think we need to run autoremove either, because we never uninstall anything and autoremove just removes unused dependencies, but happy to add it if you think it's needed.

SebastianMorawiec (approved):

PR #161, opened 2023-05-19
Internal: Fix random failures due to the warning of socket not being closed

SebastianMorawiec:
# The problem

Due to the workflow termination during socket usage, sometimes ray cleaned up stuff async, causing random failures due to the warning of socket not being closed properly

# This PR's solution

limit usage of sockets when they are not needed

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_testing/_example_wfs.py
+++ b/src/orquestra/sdk/_base/_testing/_example_wfs.py
@@ -194,6 +194,23 @@ def add_with_trigger(a, b, port, timeout: float):
     return a + b
 
 
+@sdk.task
+def infinite_task(*_):
+    import time
+
+    while True:
+        time.sleep(10)
+
+
+@sdk.workflow
+def infinite_workflow():
+    """
+    Allows reproducing scenario where tasks take some time to run.
+    This workflow is used to test termination as it will never complete
+    """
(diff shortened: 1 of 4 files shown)


src/orquestra/sdk/_base/_testing/_example_wfs.py
```diff
@@ -194,6 +194,23 @@ def add_with_trigger(a, b, port, timeout: float):
     return a + b
 
 
+@sdk.task
+def infinite_task(*_):
+    import time
+
+    while True:
```
alexjuda, line 201: **suggestion**: could you add a timeout for this loop? We probably don't want to hang forever if something goes wrong

alexjuda (approved):
Added non-blocking suggestion

src/orquestra/sdk/_base/_testing/_example_wfs.py
SebastianMorawiec: This is ray worker that is supposed to be terminated. Worst case scenario you have to restart ray. It doesnt block - thus I think it should be actually <infinite>

SebastianMorawiec: But ill add like 1hour timeout or so

PR #164, opened 2023-05-19
fix: Tasks that set n_outputs=1 in QE and in-process

jamesclark-Zapata:
# The problem

Some workflows were having issues when `n_outputs=1` was set in the task decorator.

# This PR's solution

* Only try to access a task's output by index if the output is a tuple (QE + in-process)
* Adds a test for RayRuntime

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 👩‍🔬 *Experimental*
 
 🐛 *Bug Fixes*
+* Fixed tasks that failed when explicitly state `n_outputs=1` on QE and in-process.
 
 💅 *Improvements*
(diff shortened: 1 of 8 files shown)


BenjaminMummery (approved):
LGTM, my only question is do we have tests covering the explicit single return for the other runtimes, and if not do we want to add them for consistency?

jamesclark-Zapata:
> LGTM, my only question is do we have tests covering the explicit single return for the other runtimes, and if not do we want to add them for consistency?

I think the dispatch script, Ray runtime, and in-process give us full coverage of the failure.  QE/CE runtimes are just HTTP requests around those.

PR #165, opened 2023-05-19
impr: show subprocess error for `orq up`

alexjuda:
# The problem

Sometimes, `orq up` fails because of a problem with Ray. Previously, we only told the user that starting Ray failed, but it didn't provide any insight into fixing the issue.

# This PR's solution

* Captures the subprocess stdout and stderr, and presents it to the user.
* Adds missing unit tests for `orq up` action.
* Adds integration test for `[RayManager service]-[Ray cluster]` when `ray start` fails.

# Checklist

### DIFF
--- a/tests/runtime/test_services.py
+++ b/tests/runtime/test_services.py
@@ -6,11 +6,12 @@ Unit tests for orquestra.sdk._base._services.
 """
 
 import builtins
+import subprocess
 import sys
 import typing as t
 from contextlib import contextmanager
 from pathlib import Path
-from unittest.mock import Mock
+from unittest.mock import Mock, create_autospec
 
 import pytest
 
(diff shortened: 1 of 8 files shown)


tests/runtime/test_services.py
```diff
@@ -51,6 +52,16 @@ def _fake_run(cmd, *args, **kwargs):
...
 
 
+def _mock_subprocess_module(monkeypatch):
+    mock_subprocess = Mock()
+    monkeypatch.setattr(
+        _services,
+        "subprocess",
+        mock_subprocess,
+    )
+    return mock_subprocess
```
jamesclark-Zapata, line 62: **nitpick:** ‏Any reason this is a function and not a fixture?

jamesclark-Zapata (approved):
LGTM, w/ one nitpick.

Also, `impr` isn't one of our standardised prefixes for PRs.

tests/runtime/test_services.py
alexjuda: It's used from within `_mock_subprocess_for_ray()` which is not a fixture itself because it accepts parameters. 

I think there is a way to make this work all with pytest fixtures, but the approaches I've seen make the code much more convoluted, so I opted for simple functions.

PR #168, opened 2023-05-22
chore: Add QE CLI flag

jamesclark-Zapata:
# The problem

We want to swap the defaults for logging in soon, however we don't have a QE flag.

# This PR's solution

* Add `--qe` option for logging in, mutually exclusive with `--ce`.
* Restructure CLI code to pass a `RuntimeName` instead of "is CE" bool.

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/cli/_dorq/_entry.py
+++ b/src/orquestra/sdk/_base/cli/_dorq/_entry.py
@@ -12,6 +12,8 @@ from pathlib import Path
 import click
 import cloup
 
+from orquestra.sdk.schema.configs import RemoteRuntime, RuntimeName
+
 from . import _cli_logs
 
 # Adds '-h' alias for '--help'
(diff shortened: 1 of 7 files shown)


src/orquestra/sdk/_base/cli/_dorq/_entry.py
```diff
@@ -405,20 +407,28 @@ def status():
...
     """
     Login in to remote cluster
     """
     from ._login._login import Action
 
+    runtime_name: RemoteRuntime
+    if ce:
+        runtime_name = RuntimeName.CE_REMOTE
+    else:
+        runtime_name = RuntimeName.QE_REMOTE
```
alexjuda, line 428: **nitpick**: so far we wanted to keep the `_entry` module as slim as possible, in particular devoid of any logic that would need to be tested. This ce/qe seems like a good candidate for an argument resolver. WDYT about moving it there?

jamesclark-Zapata: Seems like a lot of boilerplate, I moved it to `click`'s code via the `flag_value` kwarg in aa7d501. Does that seem better?

What we lose is the `cloup` mutually exclusive support. You can only have a single value for `runtime_name` but `orq login -s ... --ce --qe` would use `--qe` as it was the last option specified.

jamesclark-Zapata: bump

alexjuda: > What we lose is the cloup mutually exclusive support.

Yeah that's a good point. Yikes. Pre- 3facd7afdbfaf9d8878c5489443f2b4b43dbc240 seems like a better solution after all.

alexjuda (approved):

PR #187, opened 2023-05-31
refactor!: logs datastructure

alexjuda:
# The problem

There are multiple categories of logs we'll want to return. In the close future we'll implement returning venv setup logs, per task logs, and system logs. We'd like to parallelize working on the implementation, if possible.

# This PR's solution

* Changes the type returned from `WorkflowRun.get_logs()` from a [task invocation ID] ↔ [log lines] dict into a `WorkflowLogs` data structure.
* Moves the `LogReader` protocol to a separate file. I'm continuing a pattern for code organization we started recently with `orquestra.sdk._base._spaces` where we prefer to group our code by the "topic" or "concerns". Other differences like interfaces, utilities, etc are nested one level deeper.

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/abc.py
+++ b/src/orquestra/sdk/_base/abc.py
@@ -178,22 +178,6 @@ class RuntimeInterface(ABC):
         """
         raise NotImplementedError()
 
-    def get_task_logs(
-        self, wf_run_id: WorkflowRunId, task_inv_id: TaskInvocationId
-    ) -> t.List[str]:
-        """
-        See LogReader.get_task_logs()
-        """
-        raise NotImplementedError()
-
-    def get_workflow_logs(
-        self, wf_run_id: WorkflowRunId
-    ) -> t.Dict[TaskInvocationId, t.List[str]]:
-        """
-        See LogReader.get_task_logs()
-        """
-        raise NotImplementedError()
-
     @abstractmethod
     def get_workflow_project(self, wf_run_id: WorkflowRunId):
         """
--- a/src/orquestra/sdk/_base/abc.py
+++ b/src/orquestra/sdk/_base/abc.py
@@ -31,39 +31,6 @@ from orquestra.sdk.schema.workflow_run import (
 )
 
 
-class LogReader(t.Protocol):
-    """
-    A component that reads logs produced by tasks and workflows.
-    """
-
-    def get_task_logs(
-        self, wf_run_id: WorkflowRunId, task_inv_id: TaskInvocationId
-    ) -> t.List[str]:
-        """
-        Reads all available logs, specific to a single task invocation/run.
-
-        Returns:
-            Log lines printed when running this task invocation. If the task didn't
-            produce any logs this should be an empty list.
-        """
-        ...
-
-    def get_workflow_logs(
-        self, wf_run_id: WorkflowRunId
-    ) -> t.Dict[TaskInvocationId, t.List[str]]:
-        """
-        Reads all available logs printed during execution of this workflow run.
-
-        Returns:
-            A mapping with task logs. Each key-value pair corresponds to one task
-            invocation.
-            - key: task invocation ID (see
-                orquestra.sdk._base.ir.WorkflowDef.task_invocations)
-            - value: log lines from running this task invocation
-        """
-        ...
-
-
 # A typealias that hints where we expect raw artifact values.
 ArtifactValue = t.Any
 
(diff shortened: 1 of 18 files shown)


src/orquestra/sdk/_base/abc.py
```diff
@@ -15,60 +15,27 @@
...
-
+from ._logs._interfaces import LogReader, WorkflowLogs
+from ._spaces._structs import Project, ProjectRef, Workspace
 
 # A typealias that hints where we expect raw artifact values.
 ArtifactValue = t.Any
 
 
-class RuntimeInterface(ABC):
+class RuntimeInterface(ABC, LogReader):
```
alexjuda, line 38: inb4: I wanted to keep `RuntimeInterface` and `LogReader` separate but there are places where we return a single object that conforms to both interfaces, like a `foo: RuntimeInterface & LogReader`. The way to specify that in Python is to define a single class that inherits from both supertypes ([see the discussion on SO](https://stackoverflow.com/questions/60732924/how-to-specify-that-a-python-object-must-be-two-types-at-once)).

alexjuda: Instead of having a third type like `class FullRuntime(RuntimeInterface, LogReader): pass` I went with single inheritance here. I think it'll be the way to go if we change  `RuntimeInterface` into smaller interfaces specified as protocols instead of ABCs (I think that's our long term plan, right?).

tests/runtime/ray/ray_logs/test_ray_logs.py
```diff
@@ -363,7 +367,7 @@ class TestGetWorkflowLogs:
                             task_run_id=None,
                         )
                     ],
-                    {},
+                    WorkflowLogs({}, []),
```
jamesclark-Zapata, line 370: **suggestion:** ‏
```suggestion
                    WorkflowLogs(per_task={}, env_setup=[]),
```

tests/runtime/ray/ray_logs/test_ray_logs.py
```diff
@@ -379,7 +383,7 @@ class TestGetWorkflowLogs:
                             task_run_id="wf.2@inv2",
                         )
                     ],
-                    {},
+                    WorkflowLogs({}, []),
```
jamesclark-Zapata, line 386: **suggestion:** ‏
```suggestion
                    WorkflowLogs(per_task={}, env_setup=[]),
```

tests/runtime/ray/ray_logs/test_ray_logs.py
```diff
@@ -307,7 +308,7 @@ class TestGetWorkflowLogs:
         @pytest.mark.parametrize(
             "wf_run_id,parsed_logs,expected",
             [
-                pytest.param("wf.1", [], {}, id="no_parsed_logs"),
+                pytest.param("wf.1", [], WorkflowLogs({}, []), id="no_parsed_logs"),
```
jamesclark-Zapata, line 311: **suggestion:** ‏
```suggestion
                pytest.param("wf.1", [], WorkflowLogs(per_task={}, env_setup=[]), id="no_parsed_logs"),
```

tests/sdk/v2/driver/test_ce_runtime.py
```diff
@@ -1143,14 +1144,17 @@ def test_ignore_task_inv_id(
...
+        assert logs == WorkflowLogs(
+            {
+                "UNKNOWN TASK INV ID": [
+                    "<message sentinel 1>",
+                    "<message sentinel 2>",
+                    "<message sentinel 3>",
+                    "<message sentinel 4>",
+                ]
+            },
+            [],
```
jamesclark-Zapata, line 1156: **suggestion:** ‏
```suggestion
        assert logs == WorkflowLogs(
            per_task={
                "UNKNOWN TASK INV ID": [
                    "<message sentinel 1>",
                    "<message sentinel 2>",
                    "<message sentinel 3>",
                    "<message sentinel 4>",
                ]
            },
            env_setup=[],

src/orquestra/sdk/_base/abc.py
jamesclark-Zapata: I think this is OK, but we should be careful.

In this case, this just makes sure `RuntimeInterface` conforms to the `LogReader` protocol through type checking.

> The way to specify that in Python is to define a single class that inherits from both supertypes ([see the discussion on SO](https://stackoverflow.com/questions/60732924/how-to-specify-that-a-python-object-must-be-two-types-at-once)).

This is slightly different, this is about making a new type only applies for mixing two protocols to make a 3rd super-protocol. What we have above is saying `RuntimeInterface` needs to implement `LogReader` but in the SO question they're mixing two protocols.

jamesclark-Zapata (approved):

PR #191, opened 2023-06-01
imp: separate env set up logs from task logs

alexjuda:
# The problem

Users fetch can workflow logs from CE. We expose it to the user as `sdk.WorkflowRun.get_logs()`. https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/187 changed the output type to a dataclass with `per_task` and `env` fields, but all logs were put into the `per_task` category.

# This PR's solution

* Splits the logs into `per_task` and `env` categories.
* Populates the `WorkflowLogs.env` field when accessing logs from CE.
* Local runtime and QE logs are unaffected for now.

# Checklist

_Check that this PR satisfies the following items:_

- [x] Tests have been added for new features/changed behavior (if no new features have been added, check the box).
- [x] The [changelog file](CHANGELOG.md) has been updated with a user-readable description of the changes (if the change isn't visible to the user in any way, check the box).
- [x] The PR's title is prefixed with `<feat/fix/chore/imp[rovement]/int[ernal]/docs>[!]:`
- [x] The PR is linked to a JIRA ticket (if there's no suitable ticket, check the box).

PR #194, opened 2023-06-03
imp: env setup logs for local runtime

alexjuda:
~Depends on https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/191.~

https://zapatacomputing.atlassian.net/browse/ORQSDK-643

# The problem

We recently added `WorkflowLogs` data structure with `env_setup` logs. Only CE runtime filled that property, other runtimes return `[]`.

# This PR's solution

Reads local files produced by Ray and fills `.env_setup`. The approach is intended to mimic CE implementation from #191. 

I wanted to reuse as much heuristics as possible between local Ray and CE parsing. I extracted the file path rules into `_regrouping` module that's used by both runtimes. I was worried about Windows paths but I added extensive unit tests and our CI is green, so we should be fine.

To make it easier to reason about the Ray logs, I've recorded a real log dir. There's also a script for regenerating the log files. It only adds files we depend on in our logic.

# Checklist


tests/runtime/ray/ray_logs/data/ray_temp/session_2023-06-05_14-36-49_780589_86998/logs/runtime_env_agent.log
```diff
@@ -0,0 +1,12 @@
+2023-06-05 14:42:37,230	INFO runtime_env_agent.py:337 -- Creating runtime env: {"_ray_commit": "cf7a56b4b0b648c324722df7c99c168e92ff0b45", "pip": {"packages": ["polars"], "pip_check": false}} with timeout 600 seconds.
```
alexjuda, line 1: todo: add a warning about reviewing against credentials

src/orquestra/sdk/_ray/_ray_logs.py
```diff
@@ -117,34 +125,38 @@ def _get_parsed_logs(self) -> t.Iterable[WFLog]:
...
-        for log in parsed_logs:
+        for log in user_logs:
             if log.wf_run_id != wf_run_id:
                 continue
 
             if log.task_inv_id is None:
                 continue
 
-            logs_dict.setdefault(log.task_inv_id, []).append(log.json())
+            logs_dict.setdefault(log.task_inv_id, []).append(log.message)
```
alexjuda, line 157: todo: extract this change into a separate PR

tests/runtime/ray/ray_logs/data/regenerate_ray_temp.py
```diff
@@ -0,0 +1,87 @@
...
+        print(f"- {delete_path}")
+        try:
+            # 1. Assume the path is a directory.
+            shutil.rmtree(delete_path)
+        except NotADirectoryError:
+            # 2. Not a directory? It's a single file then.
+            os.remove(delete_path)
+
+
+if __name__ == "__main__":
```
SebastianMorawiec, line 86: Issue: Im not sure this should be a main function here

SebastianMorawiec: Is this file actually being used anywhere?

SebastianMorawiec (commented):
General question: Im struggling to understand how do we match log files with actual run ID we are requesting. Could you point where Can I find it to better understand context?

tests/runtime/ray/ray_logs/data/regenerate_ray_temp.py
alexjuda: The tests in [test_ray_logs.py](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/194/files#diff-3f2e64c3b741cbded3f5c4a589008728b5d8d5a48c2d609992e4dc5034e69cb2) use log files as inputs. The log files are recorded and committed in to the git repo. When I was working on this PR, regenerating and updating these log files manually was a massive PITA. `regenerate_ray_temp.py` is a dev script we can use to update the log files, e.g. after a Ray update.

alexjuda:
@SebastianMorawiec 
> General question: Im struggling to understand how do we match log files with actual run ID we are requesting. Could you point where Can I find it to better understand context?

We don't, by design!
* Matching user log lines with tasks that generated them is a different thing. See https://zapatacomputing.atlassian.net/browse/ORQSDK-840 and the following tickets.
* Matching env setup logs with tasks is not trivial. It was the reason why we return `.env_setup` as a different thing from `.per_task`. We conceived this approach for CE ([ADR](https://zapatacomputing.atlassian.net/wiki/spaces/ORQSRUN/pages/676823041/2023-04-19+CE+Logging+integration+with+SDK)); this PR follows suit for the local runtime.
* Matching env setup logs with workflow runs. Non-issue on CE, because each workflow run has a fresh `ray_temp`. Not easy to do on the local runtime. With the current implementation, if a user starts multiple workflow runs on `ray` and asks for `wf_run1.get_logs()` we'll return setup logs for all the workflow runs, not just from `wf_run1`. I can't think of a better alternative, though.

tests/runtime/ray/ray_logs/data/regenerate_ray_temp.py
SebastianMorawiec: Question: Could you move it to some different dir, like tests/runtime/ray/dev? Or somewhere basically under /dev  so its obvious its a dev script

tests/sdk/v2/logs/test_regrouping.py
```diff
@@ -0,0 +1,137 @@
...
+                Path(
+                    "/Users/emiliano/.orquestra/ray/session_latest/logs/worker-"
+                    "903b5bfa866935b17bbd3d3a64d22c7c6b746c2d522ad5fca9cf1f6a-"
+                    "01000000-249.err"
+                ),
+                id="local-absolute",
+            ),
+        ],
+    )
+    def test_valid(path: Path):
```
SebastianMorawiec, line 62: Suggestion: You mentioned the risk of Windows paths. Could you consider adding some positive and negative Windows-like paths to this UT?

tests/sdk/v2/logs/test_regrouping.py
```diff
@@ -0,0 +1,137 @@
...
+                Path(
+                    "/Users/emiliano/.orquestra/ray/"
+                    "session_2023-06-12_14-46-40_833957_38917/logs/"
+                    "runtime_env_setup-01000000.log"
+                ),
+                id="absolute-local-timestamp",
+            ),
+            pytest.param(
+                Path(
+                    "/Users/emiliano/.orquestra/ray/session_latest/logs/"
```
SebastianMorawiec, line 112: Ditto: add Windows paths

SebastianMorawiec (approved):
Approved with non-blocking suggestions

tests/runtime/ray/test_integration.py
```diff
@@ -7,6 +7,7 @@
 RuntimeInterface mocks instead of extending this file.
 """
 import json
+import re
```
BenjaminMummery, line 10: Nitpick: is `re` actually used here?

tests/runtime/ray/ray_logs/data/regenerate_ray_temp.py
BenjaminMummery: Can we specify that this is a developer-only utility in the docstring?

BenjaminMummery (approved):

tests/sdk/v2/logs/test_regrouping.py
alexjuda: I did look into Windows-specific paths in our test cases. It turns out `Path("/Users/emiliano/.orquestra/ray/session_latest/logs/")` produces `PosixPath` paths when running on UNIX-like system and `WindowsPath` when running on Windows. 

Using `WindowsPath()` directly on macOS raises an exception with a message complaining about the platform:
```
>>> import pathlib
>>> pathlib.WindowsPath("foo.txt")
  File "<stdin>", line 1, in <module>
  File "/Users/alex/.pyenv/versions/3.9.16/lib/python3.9/pathlib.py", line 1084, in __new__
    raise NotImplementedError("cannot instantiate %r on your system"
NotImplementedError: cannot instantiate 'WindowsPath' on your system
>>> 
```

Using `Path()` with a Windows-looking path string produces not the thing we'd hope for. It looks like the whole path is treated as a single file:
```
>>> pathlib.Path("C:\\foo\bar.txt")
PosixPath('C:\\foo\x08ar.txt')
>>> pathlib.Path("C:\\foo\bar.txt").parent
PosixPath('.')
```

In the end, looks like we just need to use `Path()` and run the test suite on different platforms.

PR #198, opened 2023-06-08
feat: Force stop workflows runs

jamesclark-Zapata:
# The problem

A new option was added to force stop workflow runs in CE.

# This PR's solution

* Adds a new option to our public API and CLI.
* Adds a new `State` called `KILLED`, which is the result of a force stopped workflow run.
* Adds default behaviour to `State` enums with `State.UNKNOWN`.

Example:

```
orq workflow stop --force my-wf-run-abcd123
```

or

```
run = WorkflowRun.by_id("...")
run.stop(force=True)
```

A design choice was to make `force` an optional, instead of defaulting to `False`.

This lets us make requests to CE without making a decision on if a workflow should be forcefully stopped.
Deferring the default behaviour to CE will allow this behaviour to change in the future without any changes from the client.
Of course, it's possible to pass both `True` or `False`, explicitly setting the force option. 

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_driver/_models.py
+++ b/src/orquestra/sdk/_base/_driver/_models.py
@@ -132,6 +132,12 @@ class StateResponse(str, Enum):
     SUCCEEDED = "SUCCEEDED"
     TERMINATED = "TERMINATED"
     FAILED = "FAILED"
+    KILLED = "KILLED"
+    UNKNOWN = "UNKNOWN"
+
+    @classmethod
+    def _missing_(cls, _):
+        return cls.UNKNOWN
 
 
 class RunStatusResponse(pydantic.BaseModel):
@@ -274,6 +280,15 @@ class GetWorkflowRunResponse(pydantic.BaseModel):
     """
 
     data: WorkflowRunResponse
+
+
+class TerminateWorkflowRunRequest(pydantic.BaseModel):
+    """
+    Implements:
+        https://github.com/zapatacomputing/workflow-driver/blob/873437f8157226c451220306a6ce90c80e8c8f9e/openapi/src/resources/workflow-run-terminate.yaml#L12
+    """
+
+    force: Optional[bool]
 
 
 # --- Workflow Artifacts ---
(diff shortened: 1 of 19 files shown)


src/orquestra/sdk/_base/_driver/_models.py
```diff
@@ -132,6 +132,12 @@ class StateResponse(str, Enum):
     SUCCEEDED = "SUCCEEDED"
     TERMINATED = "TERMINATED"
     FAILED = "FAILED"
+    KILLED = "KILLED"
+    UNKNOWN = "UNKNOWN"
+
+    @classmethod
+    def _missing_(cls, _):
+        return cls.UNKNOWN
```
alexjuda, line 140: **praise**: thanks for making it open-ended! TIL: [the enum `_missing_` hook](https://docs.python.org/3/library/enum.html#enum.Enum._missing_)

src/orquestra/sdk/_base/cli/_dorq/_entry.py
```diff
@@ -208,15 +207,21 @@ def wf_logs(
 @workflow.command()
 @cloup.argument("wf_run_id", required=False)
 @CONFIG_OPTION
-def stop(wf_run_id: t.Optional[str], config: t.Optional[str]):
+@cloup.option(
+    "--force/--no-force",
+    is_flag=True,
+    default=None,
+    help="Will forcefully terminate a workflow, "
+    "without waiting for it to gracefully exit.",
```
alexjuda, line 215: **suggestion**: could you describe what happens when neither `--force` nor `--no-force` is passed?

alexjuda (approved):
LGTM. Left few suggestions.

> A design choice was to make force an optional, instead of defaulting to False.
>
> This lets us make requests to CE without making a decision on if a workflow should be forcefully stopped. Deferring the default behaviour to CE will allow this behaviour to change in the future without any changes from the client. Of course, it's possible to pass both True or False, explicitly setting the force option.

Brilliant approach. Kudos.

src/orquestra/sdk/_base/cli/_dorq/_entry.py
jamesclark-Zapata: Wdyt 5537ad3?

alexjuda (approved):

PR #205, opened 2023-06-16
Fix: Fix Ray WFs failing caused by any task returning dict defined in return statement

SebastianMorawiec:
# The problem

When the task had dict defined in return statement it caused the whole workflow to fail. Caused by the fact that we treated dicts as subscribtable (with [] operator) which they are not.

# This PR's solution

Do not treat dict return values as subscribtable

# Checklist


src/orquestra/sdk/_base/_ast.py
```diff
@@ -114,9 +114,7 @@ def visit_Dict(self, node):
             # having a single `None` output.
             self.outputs.add(_AstReturnMetadata(is_subscriptable=False, n_outputs=1))
         else:
-            self.outputs.add(
-                _AstReturnMetadata(is_subscriptable=True, n_outputs=n_elements)
-            )
+            self.outputs.add(_AstReturnMetadata(is_subscriptable=False, n_outputs=1))
```
jamesclark-Zapata, line 117: **suggestion:** ‏Can you add a comment explaining why we're doing this?
Both sides of the if statement are the same, which may look odd without any explanation.

tests/runtime/ray/test_integration.py
```diff
@@ -926,3 +926,32 @@ def test_outside_workflow(self):
...
+@pytest.mark.slow
+class TestDictReturnValue:
+    """
+    Tasks that had dicts in return statement used to cause WF failures.
+    This test might look trivial bul unless AST code was majorly refactored,
+    please abstain from removing it (also refer to git history for the fix itself)
+    """
+
+    def test_dict_as_task_return_value(self, runtime: _dag.RayRuntime):
+        @sdk.task(source_import=sdk.InlineImport())
```
jamesclark-Zapata, line 940: **suggestion:** ‏
This is the default now.
```suggestion
        @sdk.task
```

jamesclark-Zapata (approved):
LGTM with some minor comments

src/orquestra/sdk/_base/_ast.py
SebastianMorawiec: Imma just colapse them into one statement for now

PR #207, opened 2023-06-16
Feat: Add new public API for WorkflowRun which returns serialized values and artifacts

SebastianMorawiec:
# The problem
We need to have public API that returns serialized results and artifacts, as this is used by Workflow-driver-ray

# This PR's solution

Add New APIs

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_api/_wf_run.py
+++ b/src/orquestra/sdk/_base/_api/_wf_run.py
@@ -392,15 +392,15 @@ class WorkflowRun:
         of these. The order is dictated by the return statement in the workflow
         function, for example `return a, b, c` means this function returns (a, b, c).
         See also:
-        https://refactored-disco-d576cb73.pages.github.io/docs/runtime/guides/workflow-syntax.html
+        https://docs.orquestra.io/docs/core/sdk/guides/workflow-syntax.html/workflow-syntax.html
 
         Args:
             wait:  whether or not to wait for workflow run completion.
-                   Uses the default options for waiting, use `wait_until_finished()` for
+                   Uses the default options for waiting, use ``wait_until_finished()`` for
                    more control.
 
         Raises:
-            WorkflowRunNotFinished: when the workflow run has not finished and `wait` is
+            WorkflowRunNotFinished: when the workflow run has not finished and ``wait`` is
                                    False
             WorkflowRunNotSucceeded: when the workflow is no longer executing, but it did not
                 succeed.
--- a/src/orquestra/sdk/_base/_api/_wf_run.py
+++ b/src/orquestra/sdk/_base/_api/_wf_run.py
@@ -400,16 +400,12 @@ class WorkflowRun:
 
         return results
 
-    def get_results_serialized(self, wait: bool = False) -> t.Sequence[t.Any]:
+    def get_results_serialized(self, wait: bool = False) -> t.Sequence[WorkflowResult]:
         """
         Retrieves workflow results in serialized form.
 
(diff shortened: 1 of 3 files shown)


src/orquestra/sdk/_base/_api/_wf_run.py
```diff
@@ -399,6 +400,38 @@ def get_results(self, wait: bool = False) -> t.Sequence[t.Any]:
 
         return results
 
+    def get_results_serialized(self, wait: bool = False) -> t.Sequence[WorkflowResult]:
```
jamesclark-Zapata, line 403: **suggestion:** ‏wdty about adding a new keyword only argument to `get_results` instead of a new method? maybe `serialized: bool = False`

SebastianMorawiec: I could do that - but Im not super into that idea. Why? Coz basically the function returns enterily different data structure. Meaning we would have function that returns either <any> (as whatever WF function reruns) OR sequence of workflowResults.

TLDR: I could do that but I prefer the way it is. maybe @alexjuda has any strong opinions?

alexjuda: I thought about a flag like this, but I think having a separate method is better. It avoids ambiguity in the returned type, usually leads to a better code completion, and avoids an `if` in our code + `assert isinstance(result, ...)` on the caller side.

jamesclark-Zapata: OK, how about using the `get_results_serialized` inside `get_results`? Just trying to minimise the number of locations we call the underlying runtime

SebastianMorawiec: Absolutely makes sense, will do that

src/orquestra/sdk/_base/_api/_wf_run.py
```diff
@@ -382,23 +379,62 @@ def get_results(self, wait: bool = False) -> t.Sequence[t.Any]:
...
+    def get_results(self, wait: bool = False) -> t.Sequence[t.Any]:
+        """
+        Retrieves workflow results, as returned by the workflow function.
+
+        A workflow function is expected to return task outputs
+        (ArtifactFutures) or constants (10, "hello", etc.). This method returns values
+        of these. The order is dictated by the return statement in the workflow
+        function, for example `return a, b, c` means this function returns (a, b, c).
+        See also:
+        https://refactored-disco-d576cb73.pages.github.io/docs/runtime/guides/workflow-syntax.html
```
alexjuda, line 395: **nitpick**: inb4 I know it's been here before. This is an internal link. Let's use the public one:
```suggestion
        https://docs.orquestra.io/docs/core/sdk/guides/workflow-syntax.html/workflow-syntax.html
```

src/orquestra/sdk/_base/_api/_wf_run.py
```diff
@@ -382,23 +379,62 @@ def get_results(self, wait: bool = False) -> t.Sequence[t.Any]:
...
+
+        Args:
+            wait:  whether or not to wait for workflow run completion.
+                   Uses the default options for waiting, use `wait_until_finished()` for
+                   more control.
+
+        Raises:
+            WorkflowRunNotFinished: when the workflow run has not finished and `wait` is
```
alexjuda, line 403: nitpick: rst syntax
```suggestion
            WorkflowRunNotFinished: when the workflow run has not finished and ``wait`` is
```

alexjuda (approved):
Added non-blocking nitpicks

PR #212, opened 2023-06-23
Feat!: Remove last bits of default config in our code

SebastianMorawiec:
# The problem

We simplified config management a while ago. Users could no longer set default config but they could still read it and sometimes it would be used on random occasions.

# This PR's solution

Remove all references to default config

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_config.py
+++ b/src/orquestra/sdk/_base/_config.py
@@ -305,15 +305,6 @@ def _resolve_prev_config_entry_for_writing(
         return EMPTY_CONFIG_FILE.configs.get(new_config_name)
 
 
-def _resolve_config_name_for_writing(config_name: str) -> str:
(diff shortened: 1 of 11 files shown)


src/orquestra/sdk/_base/_config.py
```diff
@@ -364,23 +352,25 @@ def _validate_runtime_options(
 
 
 def save_or_update(config_name, runtime_name, runtime_options):
+    if config_name in SPECIAL_CONFIG_NAME_DICT:
+        raise ValueError(f"Can't update {config_name}, it's a reserved name")
```
alexjuda, line 356: **nitpick**: this branch isn't covered by tests. Should it be? 🤷

alexjuda (approved):
<feels-good.jpg> <yells-at-defaults.png>

PR #213, opened 2023-06-23
refactor: single place for handling dates

alexjuda:
# The problem

* Python's `datetime` is full of footguns.
* I got tired of looking into my snippet library anytime I needed to do something with dates.

# This PR's solution

* Defines a single place where we create and manipulate `datetime` instants. At least the most common + problematic scenarios.
* Replaces any usage of the vanilla `datetime` module I could find with `_dates`. 

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 💅 *Improvements*
 
 🥷 *Internal*
+* Refactored `datetime` and timezone handling.
 
 📃 *Docs*
 
--- a/src/orquestra/sdk/_base/_dates.py
+++ b/src/orquestra/sdk/_base/_dates.py
@@ -1,0 +1,54 @@
+################################################################################
+# © Copyright 2023 Zapata Computing Inc.
+################################################################################
+"""
+Single place to handle datetimes and timezones without shooting yourself in the
+foot.
+"""
+
+from datetime import datetime, timezone
+
+
+def now() -> datetime:
+    """
+    Generates a timezone-aware current instant. The timezone is set to UTC.
+    """
+    return datetime.now(timezone.utc)
+
+
+def local_isoformat(instant: datetime) -> str:
+    """
+    Formats the instant using ISO8601 format with explicit time zone. The instant is
+    shifted to a local timezone for human-friendliness.
+    """
+    if instant.tzinfo is None:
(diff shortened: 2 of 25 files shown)


SebastianMorawiec (changes requested):
Question: If your intention was to move from Python Datetime to our own class, can you grep thru our codebase and remove all imports of datetime? I see even in your PR. that in_process_runtime still has it in its imports

alexjuda:
@SebastianMorawiec 
> If your intention was to move from Python Datetime to our own class

Nope. The point was *not* to have our own class, because this approach integrates badly with code we don't control. It would mean we'd have to sprinkle boilerplate whenever using or producing datetimes across our library's boundary. BTW there are [3rd-party libraries](https://arrow.readthedocs.io/en/latest/) that already do that.

The idea is to have our single facade with _functions_ that wrap operations on datetimes that are prone to errors, but without a total buy-in.

> I see even in your PR. that in_process_runtime still has it in its imports
[In-process runtime](https://github.com/zapatacomputing/orquestra-workflow-sdk/blob/87c5f20242c8fe9d1e934bf172e3a36cfe56e1f5/src/orquestra/sdk/_base/_in_process_runtime.py#L11C22-L11C30) import symbols from `datetime` module for type hint purposes only. Even if we had our own class, we'd probably still need imports like this.

alexjuda:
FYI: I've squashed commits for easier rebasing but no changes have been made after the first round of review

SebastianMorawiec:
So for lib-boundary I agree it does not make sense to produce too much of a boilerplate. For typehinting I could suggest 3 aproaches:

1.  Leave is at it is. Low-key as no effort needed
2. Add simple guard to the datetime import so its very explicity it is ONLY for type hinting:
`from typing import TYPE_CHECKING
if TYPE_CHECKING:
...`
3. Alias datetime types to our own types 1:1.

All solutions 2 and 3 improve is: future-proof that in case someone goes back here and changes stuff around datatimes it is easier to detect that something fishy is going on. (For 2nd solution you can even add explicit comment like: do not play around with datetime objects directly for the reason: <...>)

Those 2 suggestions are only non-blocking suggestions tho, so do as you wish

SebastianMorawiec (approved):
Added non-blocking suggestion in general comment section

SebastianMorawiec:
👍

PR #216, opened 2023-06-27
feat: tell the user what runtime is used when reporting successful login

BenjaminMummery:
# The problem

Under certain conditions, it is possible for a user to log in without anything in the process explicitly specifying the runtime.

Specifically, if both of the following conditions are true:
- The user does not specify a runtime flag (`--ce` or `--qe`).
- No stored config fir this URL exists, or the stored config uses CE

the process looks like this:

```
 $ orq login -s my_cluster_url
Continue the login process in your web browser. Press [ctrl-c] to cancel.
Token saved in config file.
Configuration name for my_cluster_url is my_config_name
```

(the same is true for ` $ orq login -c my_config_name`)

In this case, we rely entirely on the user being cognisant of the fact that CE is the default to tell them to what runtime `my_config_name` relates.

# This PR's solution

Adds runtime information to the final message:

```
 $ orq login -s my_cluster_url
Continue the login process in your web browser. Press [ctrl-c] to cancel.
Token saved in config file.
Configuration name for my_cluster_url with runtime CE_REMOTE is 'my_config_name'.
```

I also made a few small related improvements:
- Instructions for manual login will now always include the appropriate runtime flag. As written this would only show the `--ce` flag, and so was outdated by the change in default from QE to CE. The new version is agnostic to the default.
- The saved config name is now shown in quotes, helping it stand out when scanning across the line.

# Checklist


alexjuda (approved):
❤️

PR #219, opened 2023-06-28
Int: Prepare CE Runtime to use different client URI's

SebastianMorawiec:
# The problem

Our architecture has hard-coded way of providing URIs. We have strong dependency on how we combine base uri with endpoint.

# This PR's solution

Decouple CE_CLIENT with how we provide URIs to communicate. 

# Checklist


src/orquestra/sdk/_base/_driver/_client.py
```diff
@@ -31,33 +31,44 @@
...
+        "get_login_url": "/api/login",
+        # Workspaces
+        "list_workspaces": "/api/catalog/workspaces",
+        "list_projects": "/api/catalog/workspaces/{}/projects",
+    }
+
+    def __init__(self, base_uri):
+        self._base_uri = base_uri
+
+    def provide_uri(self, action, parameters: Optional[Tuple[str, ...]] = None) -> str:
```
alexjuda, line 67:
```suggestion
    def uri_for(self, action_id: str, parameters: Optional[Tuple[str, ...]] = None) -> str:
```

src/orquestra/sdk/_base/_driver/_client.py
```diff
@@ -590,7 +611,10 @@ def get_workflow_run_result(
         """
 
         resp = self._get(
-            API_ACTIONS["get_workflow_run_result"].format(result_id), query_params=None
+            self._uri_provider.uri_for(
+                "get_workflow_run_result", parameters=(result_id,)
+            ),
+            query_params=None,
```
alexjuda, line 617: **non-blocking,question**: does the addition of `query_params=None` change anything here?

tests/runtime/ray/test_integration.py
```diff
@@ -37,9 +37,16 @@ def runtime(
     config = configs.RuntimeConfiguration(
         config_name="test-config",
         runtime_name=configs.RuntimeName.RAY_LOCAL,
+        runtime_options={
+            "address": None,
+            "log_to_driver": True,
+            "storage": None,
+            "temp_dir": None,
+            "configure_logging": None,
+        },
```
alexjuda, line 46: **question**: ray configuration option in tests is a super delicate topic. What change does this diff make?

alexjuda (dismissed):
Added non-blocking comments. Blocking until the PR checklist is filled.

tests/runtime/ray/test_integration.py
SebastianMorawiec: Previously we used those runtime_options is from_runtime_configuration. Now as I removed this function I added its logic to constructor, which is used in our tests. Ill double check all usages of those params if this might ahve any impact on anything

src/orquestra/sdk/_base/_driver/_client.py
SebastianMorawiec: No, but its required argument. You can see it was there before, just in another line

alexjuda: oooh, right.

tests/runtime/ray/test_integration.py
alexjuda: **suggestion**: it's very easy to misconfigure Ray, making the tests unreliable. After these changes the options will be defined in multiple places, which increases the risk of silent errors. Could you use a single place where we define the runtime options for Ray config used in tests (SSOT)?

SebastianMorawiec: Good point. Changed all definitions of this config to point out to our global config we use when user uses "local"

src/orquestra/sdk/_base/_driver/_client.py
```diff
@@ -119,33 +132,33 @@ class DriverClient:
     Client for interacting with the Workflow Driver API via HTTP.
     """
 
-    def __init__(self, base_uri: str, session: requests.Session):
-        self._base_uri = base_uri
+    def __init__(self, session: requests.Session, uri_provider: ExternalUriProvider):
+        self._uri_provider = uri_provider
```
jamesclark-Zapata, line 136: **suggestion:** For ‏Either now or next PR:
We want to be able to pass any `URIProvider`, so this should be a protocol or something.

src/orquestra/sdk/_base/_driver/_client.py
```diff
@@ -119,33 +132,33 @@ class DriverClient:
...
         self._session = session
 
     @classmethod
-    def from_token(cls, base_uri: str, token: str):
+    def from_token(cls, token: str, uri_provider: ExternalUriProvider):
```
jamesclark-Zapata, line 140: **suggestion:** ‏same here

src/orquestra/sdk/_ray/_dag.py
```diff
@@ -187,23 +187,14 @@ class RayParams:
 class RayRuntime(RuntimeInterface):
     def __init__(
         self,
-        client: RayClient,
         config: RuntimeConfiguration,
         project_dir: Path,
+        client=None,
```
jamesclark-Zapata, line 192: **nitpick:** ‏Can you add type annotations for `client`?

src/orquestra/sdk/_ray/_dag.py
```diff
@@ -187,23 +187,14 @@ class RayParams:
...
-        self._log_reader: LogReader = _ray_logs.DirectRayReader(
-            _services.ray_temp_path()
-        )
-
-    @classmethod
-    def from_runtime_configuration(
-        cls, project_dir: Path, config: RuntimeConfiguration, verbose: bool
-    ) -> "RuntimeInterface":
-        client = RayClient()
+        self._client = client
```
jamesclark-Zapata, line 197: **suggestion:** ‏This mutates the input `client`. I think it's OK here, but this might also work:
```suggestion
        self._client = client or RayClient()
```

tests/runtime/ray/test_dag.py
```diff
@@ -121,11 +125,9 @@ class TestRayRuntime:
 
     @staticmethod
     @pytest.fixture
-    def runtime_config():
-        return RuntimeConfiguration(
-            config_name="TestRayRuntime",
-            runtime_name=RuntimeName.RAY_LOCAL,
-        )
+    def runtime_config(monkeypatch):
+        monkeypatch.setattr(_dag.RayRuntime, "startup", lambda *_, **__: ...)
```
jamesclark-Zapata, line 129: **thought:** ‏It might be a bit unexpected for a fixture that returns a config to also mock the startup method.

src/orquestra/sdk/_base/_factory.py
```diff
@@ -22,24 +22,49 @@ class to use.
...
+
+    client = orquestra.sdk._base._driver._client.DriverClient.from_token(
+        token=token, uri_provider=uri_provider
+    )
+
+    return orquestra.sdk._base._driver._ce_runtime.CERuntime(
         config=config,
+        client=client,
         verbose=verbose,
     )
```
jamesclark-Zapata, line 70: **issue:** ‏We don't have any unit tests for this

src/orquestra/sdk/_base/_driver/_client.py
SebastianMorawiec: We just though of either protocol in this file and implementations is separate, or just Union type. Either we thought to leave it up for the actual implementation 👍

tests/runtime/ray/test_dag.py
SebastianMorawiec: good idea, moved it to separate fixture

jamesclark-Zapata (approved):

PR #221, opened 2023-06-29
 feat: api get system logs from local ray

BenjaminMummery:
ORQSDK-865

# The problem

At present, `wf.get_logs().system` will always be an empty list for local ray workflow runs.

# This PR's solution

There isn't a hard-and-fast rule for what constitutes a 'system' log in this context. There also isn't a mandate for providing system logs via the API since the log files exist locally on the user's machine. Therefore we instead insert an artificial log line directing the user to the appropriate directory.

```
>>> wf.get_logs().system
["WARNING: we don't parse system logs for the local runtime. The log files can be found in the directory '~/.orquestra/ray/session_latest/logs."]
```

(We also insert an almost identical message for the 'other' logs)

# Checklist

### DIFF
--- a/src/orquestra/sdk/_ray/_ray_logs.py
+++ b/src/orquestra/sdk/_ray/_ray_logs.py
@@ -138,7 +138,7 @@ class DirectRayReader:
         system_warning = (
             "WARNING: we don't parse system logs for the local runtime. "
             "The log files can be found in the directory "
-            "'~/.orquestra/ray/session_latest/logs."
+            f"'{self._ray_temp}'"
         )
         return [system_warning]
 
@@ -146,7 +146,7 @@ class DirectRayReader:
         other_warning = (
             "WARNING: we don't parse uncategorized logs for the local runtime. "
             "The log files can be found in the directory "
-            "'~/.orquestra/ray/session_latest/logs."
+            f"'{self._ray_temp}'"
         )
         return [other_warning]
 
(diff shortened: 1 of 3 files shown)


src/orquestra/sdk/_ray/_ray_logs.py
```diff
@@ -130,6 +130,26 @@ def _get_env_setup_lines(self) -> t.Sequence[str]:
...
+            "The log files can be found in the directory "
+            "'~/.orquestra/ray/session_latest/logs."
+        )
+        return [system_warning]
+
+    def _get_other_log_lines(self) -> t.Sequence[str]:
+        other_warning = (
+            "WARNING: we don't parse uncategorized logs for the local runtime. "
+            "The log files can be found in the directory "
+            "'~/.orquestra/ray/session_latest/logs."
```
alexjuda, line 149: **non-blocking**: WDYT of using the [ray_temp()](https://github.com/zapatacomputing/orquestra-workflow-sdk/blob/b25507680dd01004aef153961d80bebc62392af1/src/orquestra/sdk/_base/_services.py#L34C5-L34C18) helper, as a SSOT? This would prevent us from proving bad advice in scenarios where the RAY_TEMP is in a different location. Can't think of a valid use case today, but it might save our butts in the future.

alexjuda (approved):
Added non-blocking suggestion

src/orquestra/sdk/_ray/_ray_logs.py
BenjaminMummery: Good idea, done in [0ed1c25](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/221/commits/0ed1c25c45dc02bc4ca7f21681735cb82f7b5196)

PR #222, opened 2023-06-30
Feat!: Changed wf_def.get_tasks() return type from set, to topologically sorted list

SebastianMorawiec:
# The problem

get_tasks() was returning `set()` object which was confusing for a lot of users
https://zapatacomputing.atlassian.net/browse/ORQSDK-885

# This PR's solution

Change returned type to list, which is sorted topologically. 

# Checklist


src/orquestra/sdk/_base/_graphs.py
```diff
@@ -50,15 +55,15 @@ def topological_sort(graph_to_sort: Graph) -> t.List[Node]:
...
         n_deps = list(graph.get(n, []))
         for m in n_deps:
-            graph[n].remove(m)
-            inverted_graph.get(m, set()).remove(n)
+            graph[n].pop(m)
+            inverted_graph.get(m, dict()).pop(n)
 
             if not inverted_graph.get(m):
-                S.add(m)
+                S[m] = None
```
alexjuda, line 66: **suggestion**: after this PR we'll be dealing with `dict`s with `None` as values. `d.get("k")` will return `None` even if the value's there. Changing the condition into an explicit form might be less bug-prone.

```
if m not in inverted_graph:
```
or
```
if len(inverted_graph.get(m, dict())) == 0:
```

SebastianMorawiec: inverted_graph is dict of {node, dict} (was {node, set} before my changes). This checks if inverted_graph(m) == {}. "None" object ive added to internal dict has nothing to do with this line of code.

jamesclark-Zapata (approved):
As discussed, I think the updated check in `topological_sort` is OK, but we should be careful and check for any bug reports

CHANGELOG.md
```diff
@@ -6,11 +6,12 @@
 * Removed `RuntimeConfig.load_default()`
 * Removed any support for default configuration
 * `sdk.secret` functions will no longer use default configuration from local runtimes. Config has to be passed explicitly unless running on remote cluster
-* Add `--list` option to `orq login` that displays the stored remote logins, which runtimes they are using, and whether their access tokens are up to date.
+* `WorkflowDef.get_tasks()` now returns topologically sorted list of TaskRuns instead of set
```
jamesclark-Zapata, line 10: Watch out for this changelog modification!

PR #225, opened 2023-07-04
chore: Use NamedTuple when returning IDs

jamesclark-Zapata:
# The problem

We returned a tuple of `(WorkflowRunId, Optional[TaskInvocationId]`, Optional[TaskRunId])` which can be a source of bugs. Accessing a tuple by index means it is easier to make copy/paste errors.

# This PR's solution

Changes this Tuple to a `NamedTuple` so one can access the different run IDs by attribute. For example:

Old:
```python
my_task_inv_id = sdk.current_run_ids()[1]
```

New:
```python
my_task_inv_id = sdk.current_run_ids().task_invocation_id
```

As this is a `NamedTuple`, this doesn't change any old behaviour and only adds a nicer way to get these IDs.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 🐛 *Bug Fixes*
 
 💅 *Improvements*
+* `sdk.current_run_ids()` now returns a `NamedTuple` called `CurrentRunIDs` to help with typing.
 
 🥷 *Internal*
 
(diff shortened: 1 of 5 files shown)


SebastianMorawiec (approved):
Be careful with changelog before merge. There was release in between.
Also: Changing tuple to namedtuple might be considered breaking change, but I think for this usecase its fine.

SebastianMorawiec (changes requested):
Oh Ive just noticed: there is no tests for that? Or the tests are assuming its just a tuple? Can we adjust tests and/or add to check if namedTuple is created correctly?

SebastianMorawiec (approved):
Remember to update the changelog. Currenlty entry is added to wrong release

PR #231, opened 2023-07-11
fix: Use non-root user to install Python packages and use a venv

emre-aydin:
# The problem

Installing Python modules as root user cause custom images to fail.

# This PR's solution

Installing Python modules as a non-root user and in a virtualenv to avoid edge cases.

# Jira Ticket

https://zapatacomputing.atlassian.net/browse/OP-166

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 👩‍🔬 *Experimental*
 
 🐛 *Bug Fixes*
+* Install Python modules in a venv using a non-root user to fix errors in custom Docker images.
 
 💅 *Improvements*
 * `sdk.current_run_ids()` now returns a `NamedTuple` called `CurrentRunIDs` to help with typing.
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -21,13 +21,20 @@ fi
 
 rm -rf /var/lib/apt/lists/*
 
+useradd -ms /bin/bash -d /home/orquestra orquestra --uid 1000 --gid 100
+EOF
+
+USER 1000
+WORKDIR /home/orquestra
+
+ENV VIRTUAL_ENV=/home/orquestra/venv
+RUN python -m venv "$VIRTUAL_ENV"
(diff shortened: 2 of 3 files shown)


jamesclark-Zapata (commented):
Does this require any updates to custom image docs?

emre-aydin:
> Also, what about the CUDA image?

I will send that in a separate PR.

emre-aydin:
> Does this require any updates to custom image docs?

Yes, good point. I forgot you had the docs in the same repo. I was planning to submit a separate PR to a docs repo for that. Let me add it here.

emre-aydin:
@jamesclark-Zapata having a look at the current docs, we don't mention anywhere that we install the Python modules as `root` user. Do you think it would be useful to mention that we now use a virtualenv and install using `orquestra` user? Since we extend `PATH` variable, all pip installs will use the virtualenv anyway so I'm not sure if we need to mention it in the docs.

jamesclark-Zapata:
> @jamesclark-Zapata having a look at the current docs, we don't mention anywhere that we install the Python modules as `root` user. Do you think it would be useful to mention that we now use a virtualenv and install using `orquestra` user? Since we extend `PATH` variable, all pip installs will use the virtualenv anyway so I'm not sure if we need to mention it in the docs.

It might be worthwhile mentioning it in the "Building an Image" section, we already mention how to install Ubuntu packages, for example.
Virtual environments mean the Julia/Python integrations need more configuration.

emre-aydin:
> > @jamesclark-Zapata having a look at the current docs, we don't mention anywhere that we install the Python modules as `root` user. Do you think it would be useful to mention that we now use a virtualenv and install using `orquestra` user? Since we extend `PATH` variable, all pip installs will use the virtualenv anyway so I'm not sure if we need to mention it in the docs.
> 
> It might be worthwhile mentioning it in the "Building an Image" section, we already mention how to install Ubuntu packages, for example. Virtual environments mean the Julia/Python integrations need more configuration.

@jamesclark-Zapata updated the docs.

jamesclark-Zapata (approved):

PR #233, opened 2023-07-11
fix: move venv for studio

roryschadler:
# The problem

Studio overwrites anything stored under `/home/orquestra` on runtime, with the user's persistent volume. ([OP-179](https://zapatacomputing.atlassian.net/browse/OP-179))

# This PR's solution

I moved the new `venv` under `/opt/orquestra` after discussion with @jamesclark-Zapata.

# Checklist

[OP-179]: https://zapatacomputing.atlassian.net/browse/OP-179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docker/Dockerfile
```diff
@@ -22,13 +22,16 @@ fi
...
+chown -R 1000:100 /opt/orquestra
 EOF
 
 USER 1000
 WORKDIR /home/orquestra
 
-ENV VIRTUAL_ENV=/home/orquestra/venv
-RUN python -m venv "$VIRTUAL_ENV"
+ENV VIRTUAL_ENV=/opt/orquestra/venv
+RUN python -m venv --system-site-packages "$VIRTUAL_ENV"
```
alexjuda, line 34: **question**: I don't usually use the `--system-site-packages` option, which makes my venvs fully isolated apart from Python's standard library. What's the backstory for this change?

emre-aydin:
@roryschadler I'm currently working on updating the cuda image in the same way using virtualenv. Do we have the same problem there? Should I have the venv in `/opt/orquestra/venv` in it as well?

docker/Dockerfile
jamesclark-Zapata: If users want to install packages in their home folder (the home folder is persisted), the venv needs access to the user site packages.
There's no option to use only the user site packages.
I'm not sure we've done enough investigating the other consequences to make this change.
For example, can I uninstall packages OK, will the system site packages cause an issue, etc.

alexjuda (approved):

PR #235, opened 2023-07-12
fix: Use non-root user to install Python packages and use a venv in CUDA Docker image

emre-aydin:
# The problem

Root user is used for installing Python module and the system Python executable is used, which creates problems for custom image authors.

# This PR's solution

This is a follow-up PR to another and makes the same change for the CUDA images.

# Jira ticket

https://zapatacomputing.atlassian.net/browse/OP-166

# Checklist

### DIFF
--- a/docker/cuda.Dockerfile
+++ b/docker/cuda.Dockerfile
@@ -39,26 +39,40 @@ python -m pip install pybind11
 apt-get install --yes git
 export DEBIAN_FRONTEND=noninteractive
 apt-get install --yes cmake
+
 rm -rf /var/lib/apt/lists/*
+
+useradd -ms /bin/bash -d /home/orquestra orquestra --uid 1000 --gid 100
+
+mkdir -p /opt/orquestra
+chown -R 1000:100 /opt/orquestra
 EOF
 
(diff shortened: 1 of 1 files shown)


docker/cuda.Dockerfile
```diff
@@ -35,25 +36,30 @@ ENV CUQUANTUM_DIR=/opt/nvidia/cuquantum
 # get required tools to build qsim
 RUN <<EOF
 python -m pip install pybind11
```
alexjuda, line 38: It looks like this `pip install` happens outside of our venv. Should this line be moved down?

emre-aydin: Yes, AFAIU it's required only for that install so I chose to not pollute our venv with it (I might be mistaken though).

emre-aydin: @alexjuda after adapting to changes in https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/233, this is now being installed in the same virtualenv.

roryschadler (approved):

alexjuda (approved):
I'll rely on @roryschadler's approval wrt. alignment with the other docker images discussed this week! Apart from it, LGTM

PR #246, opened 2023-07-17
Feat: Verify if workflow resources are sufficient at submission time

SebastianMorawiec:
# The problem

Workflows fail very late when there are no workers to run given task
https://zapatacomputing.atlassian.net/browse/ORQSDK-908

# This PR's solution

Error out when workflow resources dont match task resources

# Checklist


src/orquestra/sdk/_base/_driver/_ce_runtime.py
```diff
@@ -52,6 +52,37 @@ def _get_max_resources(workflow_def: WorkflowDef) -> _models.Resources:
...
 
 
+def _verify_workflow_resources(
+    wf_resources: _models.Resources, max_resources: _models.Resources
+):
+    # Default workflow CPU and memory defined on platform side, if Nones are passed
+    # https://github.com/zapatacomputing/workflow-driver/blob/1d4d0552e44cfd11238b823cbf66a33d4d2e8593/pkg/config/config.go#L22  # noqa
+    default_workflow_cpu = "2"
+    default_workflow_memory = "2Gi"
+    insufficient_resources = set()
```
alexjuda, line 62: **suggestion**: looks like the only usage of `insufficient_resources` is to construct the exception message. `insufficient_resources` is a `set`, so the `str()` will differ between Python processes. `list` would make the messages deterministic without affecting the logic:
```suggestion
    insufficient_resources = list()
```

alexjuda (changes requested):
Non-blocking nitpick added.

Blocking: please edit the changelog.

alexjuda (approved):

PR #253, opened 2023-07-21
feat: raise informative error message when depickling fails due to python version mismatch.

BenjaminMummery:
[ORQSDK-844]

# The problem

CE runs on python 3.9. Differing versions of python between the pickler and unpickler can cause a SystemError. Therefore if the local version of python is not 3.9, the following error can appear:

```
  File "/home/orquestra/venv/lib/python3.9/site-packages/orquestra/sdk/_ray/_build_workflow.py", line 194, in _ray_remote
    wrapped_return = wrapped(*inner_args, **inner_kwargs)
  File "/tmp/ray/session_2023-05-15_09-25-45_300259_8/runtime_resources/pip/1e7cfcc6d6350e4a83bef0c9a0220b918d2795b3/virtualenv/lib/python3.9/site-packages/orquestra/sdk/_ray/_build_workflow.py", line 139, in __call__
    return self._user_fn(*args, **kwargs)
  File "/Users/sebastianmorawiec/git/evangelism-workflows/demos/testing_sdk/mlflow_demo.py", line 106, in fake_model_training
SystemError: unknown opcode
```

This does not actually tell the user what's caused the problem or how to fix it.

# This PR's solution

Wraps the unpickling in a try-except block to raise a more helpful error:

```
EnvironmentError: Could not deserialise data. This may be due to a python version mismatch. Please try updating your python version.
```

# Checklist

- [ ] Tests have been added for new features/changed behavior (if no new features have been added, check the box).

[ORQSDK-844]: https://zapatacomputing.atlassian.net/browse/ORQSDK-844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -214,7 +214,14 @@ def _ray_remote(*inner_args, **inner_kwargs):
                     deserialize=serialization,
                 )
 
-                wrapped_return = wrapped(*inner_args, **inner_kwargs)
+                try:
+                    wrapped_return = wrapped(*inner_args, **inner_kwargs)
+                except SystemError as e:
```
jamesclark-Zapata, line 219: **question:** ‏Where does `SystemError` come from? Let's document what raises this exception.

src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -214,7 +214,14 @@ def _ray_remote(*inner_args, **inner_kwargs):
...
+                    raise EnvironmentError(
```
jamesclark-Zapata, line 220: **nitpick:** ‏`EnvrionmentError` is only available for backwards compatibility (since 3.3) and is an alias of `OSError`. Perhaps a custom exception may be a better thing to use.

src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -214,7 +214,14 @@ def _ray_remote(*inner_args, **inner_kwargs):
...
+                        "Could not deserialise data. "
+                        "This may be due to a python version mismatch. "
+                        "Please try updating your python version."
```
jamesclark-Zapata, line 223:
```suggestion
                        "This may be due to a Python version mismatch. "
                        "Please try updating your Python version."
```

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@
 🐛 *Bug Fixes*
 
 💅 *Improvements*
+* Added an informative error for situations where a mismatch in python versions between CE and the SDK cause deserialisation to fail.
```
jamesclark-Zapata, line 19: **suggestion:** ‏
```suggestion
* Added an informative error for situations where a mismatch in Python versions between CE and the SDK cause deserialisation to fail.
```

jamesclark-Zapata (commented):
We have the submitted Python version available in the IR. Can we use this information to confirm that the Python version is a possible culprit?

src/orquestra/sdk/_ray/_build_workflow.py
jamesclark-Zapata: Actually, the action here isn't too helpful. Can we tell the user what version of Python they should install?

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 * `KILLED` workflows will no longer show an error when waiting for the workflow to complete.
 
 💅 *Improvements*
+* Added an informative error for situations where a mismatch in Python versions between CE and the SDK cause deserialisation to fail.
```
jamesclark-Zapata, line 24:
```suggestion
* Added an informative error for situations where a mismatch in Python versions between your local environment and the remote cluster prevents tasks from executing.
```

src/orquestra/sdk/_base/_driver/_ce_runtime.py
```diff
@@ -32,6 +37,10 @@
 
 from . import _client, _exceptions, _models
 
+CE_REQUIRES_PYTHON_VERSION: Version = Version(
+    original="3.9", major=3, minor=9, patch=0, is_prerelease=False
+)
```
jamesclark-Zapata, line 42: **issue:** ‏CE actually uses 3.9.17. How will we make sure this is updated in the future?

src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -152,7 +157,10 @@ def __call__(self, *wrapped_args, **wrapped_kwargs):
         for name, kwarg in wrapped_kwargs.items():
             kwargs[name] = self._unpack_argument(kwarg, name)
 
-        return self._user_fn(*args, **kwargs)
+        try:
+            return self._user_fn(*args, **kwargs)
+        except SystemError as e:
+            raise exceptions.TaskWrappingError from e
```
jamesclark-Zapata, line 163: **thought:** This will catch different op-codes, but are there any other error cases we should think about? We'll see I guess!

src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -239,7 +259,10 @@ def _ray_remote(*inner_args, **inner_kwargs):
                     unpacked=unpacked,
                 )
 
-    named_remote = client.add_options(_ray_remote, **ray_options)
+    try:
+        named_remote = client.add_options(_ray_remote, **ray_options)
+    except exceptions.TaskWrappingError:
+        raise
```
jamesclark-Zapata, line 265: **question:** ‏Does the underlying Ray function get called here? I didn't expect `.add_options()` to raise this exception.

tests/runtime/ray/test_build_workflow.py
```diff
@@ -292,22 +292,23 @@ def test_handles_taskwrappingerror(
...
+        with pytest.raises(expected_exception):  # type: ignore
             _ = _build_workflow.make_ray_dag(client, workflow, wf_run_id, None)
 
     @staticmethod
     def test_pythonversionmismatch_message(monkeypatch: pytest.MonkeyPatch):
         # Given
         workflow = workflow_parametrised_with_resources().model
 
-        workflow.metadata.python_version = ir.Version(
+        workflow.metadata.python_version = ir.Version(  # type: ignore
```
jamesclark-Zapata: **suggestion:** ‏`# type: ignore` should only be used as a last resort.

```suggestion
        assert workflow.metadata is not None
        workflow.metadata.python_version = ir.Version(  # type: ignore
```

tests/runtime/ray/test_build_workflow.py
```diff
@@ -292,22 +292,23 @@ def test_handles_taskwrappingerror(
...
+        assert workflow.metadata is not None
         workflow.metadata.python_version = python_version
         client = create_autospec(_build_workflow.RayClient)
         client.add_options = Mock(side_effect=exceptions.TaskWrappingError)
 
         wf_run_id = "<wf run id sentinel>"
 
         # When
-        with pytest.raises(expected_exception):
+        with pytest.raises(expected_exception):  # type: ignore
```
jamesclark-Zapata: **nitpick:** ‏The reason you're seeing a type error here is because `expected_exception` is typed as an `Exception` (which is an instance of an exception) but what you actually pass is `Type[Exception]`.

I'd remove `# type: ignore` and change `expected_exception: Exception` to `expected_exception: Type[Exception]`. (Type is from `typing`)

PR #264, opened 2023-07-28
Feat: Remove all instances of project prompts for listing workflows 

SebastianMorawiec:
# The problem

Platform ignored project parameter when listing workflows. We passed it causing confusion
https://zapatacomputing.atlassian.net/browse/ORQSDK-874

# This PR's solution

_Please describe the PR's content. Bonus points for code examples, error messages, screenshots._

# Checklist


src/orquestra/sdk/_base/_spaces/_structs.py
```diff
@@ -2,14 +2,15 @@
...
 
 from orquestra.sdk.schema.workflow_run import ProjectId, WorkspaceId
 
 
 @dataclass(frozen=True)
 class ProjectRef:
-    workspace_id: WorkspaceId
-    project_id: ProjectId
+    workspace_id: Optional[WorkspaceId]
+    project_id: Optional[ProjectId]
```
alexjuda, line 13: **issue**: the original reason why we introduced the `ProjectRef` struct was to encode situations where we need both `workspace_id` and `project_id`, specifically to avoid situations where both can be set to `None` thanks to mypy. This diff goes against the original design. What's the motivation behind doing it?

I haven't dug deeper into it yet, but it seems like instead of making these fields `Optional` we should just change the `foo: ProjectRef` type annotations to `foo: WorkspaceId`.

src/orquestra/sdk/_base/cli/_dorq/_repos.py
```diff
@@ -61,21 +61,20 @@ def get_config_name_by_run_id(self, wf_run_id: WorkflowRunId) -> ConfigName:
...
-        ]
+    # def list_wf_run_ids(
+    #     self, config: ConfigName, project: ProjectRef
+    # ) -> t.Sequence[WorkflowRunId]:
+    #     return [
+    #         run.id
+    #         for run in self.list_wf_runs(
+    #             config, project.workspace_id
+    #         )
+    #     ]
```
alexjuda, line 72:
```suggestion
```

alexjuda (changes requested):

alexjuda (approved):

PR #266, opened 2023-07-28
docs: MLflow connection

alexjuda:
# The problem

Using MLflow tracking servers deployed in an Orquestra cluster requires some manual set up. We didn't have the docs for it.

# This PR's solution

Extends the MLflow guide we already had with sections about:
* How to enable an MLflow tracking server instance via Portal.
* The usage of `sdk.mlflow.get_tracking_uri()` and `sdk.mlflow.get_tracking_token()`.
* The lack of global state sharing across task runs.

# Checklist


docs/guides/mlflow.rst
```diff
@@ -1,18 +1,69 @@
...
+         This global state leakage can happen if two tasks have the same ``@sdk.task(...)`` decorator arguments and the Ray scheduler decides to reuse the same worker process, or if you're using the ``in_process`` runtime.
+         It's best to assume the global state set by one task run won't be available to other task runs.
 
 Uploading Artifacts
 -------------------
 
-Artifacts must be written to a file before being uploaded to MLflow. Rather than manually configuring an appropriate temporary directory for creating these files, we provide the ``get_temp_artifacts_dir()`` utility. This function can be used in any runtime and will return an appropriate directory.
+MLflow artifacts must be written to a file before being uploaded to an MLflow tracking server.
+The problem is: it's not always obvious what's a suitable local directory for writing such files, especially on the remote runtime.
+The Workflow SDK provides a ``get_temp_artifacts_dir()`` utility.
```
BenjaminMummery, line 57: Suggestion (non-blocking): how do you feel about the following to make the causal relationship here clearer?
```suggestion
The problem is: it's not always obvious what's a suitable local directory for writing such files, especially on the remote runtime.
In order to address this, the Workflow SDK provides a ``get_temp_artifacts_dir()`` utility.
```

docs/guides/mlflow.rst
```diff
@@ -1,18 +1,69 @@
...
+.. [#f1] Global state can leak between task runs if subsequent task runs are executed in the same Python process.
+         This global state leakage can happen if two tasks have the same ``@sdk.task(...)`` decorator arguments and the Ray scheduler decides to reuse the same worker process, or if you're using the ``in_process`` runtime.
+         It's best to assume the global state set by one task run won't be available to other task runs.
 
 Uploading Artifacts
 -------------------
 
-Artifacts must be written to a file before being uploaded to MLflow. Rather than manually configuring an appropriate temporary directory for creating these files, we provide the ``get_temp_artifacts_dir()`` utility. This function can be used in any runtime and will return an appropriate directory.
+MLflow artifacts must be written to a file before being uploaded to an MLflow tracking server.
+The problem is: it's not always obvious what's a suitable local directory for writing such files, especially on the remote runtime.
```
BenjaminMummery, line 56: Suggestion (non-blocking): I think this has slightly better clarity and flow, wdyt?
```suggestion
However, it's not always obvious what is and is not a suitable local directory for writing such files, especially on the remote runtime.
```

setup.cfg
```diff
@@ -88,6 +88,8 @@ dev =
     pyright
     pytest-dependency
     pytest_httpserver
+    # Required for testing snippets in docs.
+    mlflow-skinny
```
BenjaminMummery, line 92: Suggestion: can we do it this way so that any automatic sorting won't separate the dependency from the comment?
```suggestion
    mlflow-skinny  # Required for testing snippets in docs.
```

BenjaminMummery (changes requested):
Nice guide! I had a few minor fixes and some suggestions :)

docs/examples/tests/test_mlflow_utilities.py
```diff
@@ -6,48 +6,255 @@
...
+            # Then
+            assert mocked_environ["MLFLOW_TRACKING_TOKEN"] == token
+            assert (
+                mocked_environ["MLFLOW_TRACKING_URI"]
+                == f"http://{cr_name}.{ws_namespace}:{mlflow_port}"
+            )
+
+    class TestTrackingTask:
+        @staticmethod
+        def test_local(mocked_environ, monkeypatch):
```
BenjaminMummery, line 218: This test fails if config.json doesn't exists. There's a fixture in the main tests' conftest.py that mocks the config file location, we should do something similar here

BenjaminMummery (changes requested):
Nice guide! I had a few minor fixes and some suggestions :)

docs/guides/mlflow.rst
```diff
@@ -1,18 +1,69 @@
-.. © Copyright 2023 Zapata Computing Inc.
+MLflow In Orquestra
```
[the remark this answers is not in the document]
alexjuda: I'm in favor of being explicit, because this document is about integrating with MLflow deployed to Orquestra clusters and not MLflow in general.

docs/guides/mlflow.rst
```diff
@@ -1,18 +1,69 @@
...
+
+
+.. image:: images/portal-mlflow.png
+   :align: center
+
+
+Connecting From Python
+----------------------
+
+MLflow provides their own `Python client library <https://pypi.org/project/mlflow/>`_.
```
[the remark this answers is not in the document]
alexjuda: Thanks, by accident I assumed MLflow was a person with they/them pronouns 😄 power of habit

docs/guides/mlflow.rst
```diff
@@ -1,18 +1,69 @@
...
+
+When used on a local machine, the utility functions make use of the cluster URI and token stored in the config you created when :doc:`logging in<../tutorials/remote>`.
+
+When used in Orquestra Studio or within a task running on CE, the utility functions ignore the ``cluster_name=`` argument, and use a different mechanism for cluster-local URIs and auth.
+
+Passing in ``workspace_id`` to ``get_tracking_uri()`` is required regardless of the execution environment.
+
+
+Connecting From A Task
+----------------------
```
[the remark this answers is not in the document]
alexjuda: Went with option 1, thanks!

docs/guides/mlflow.rst
```diff
@@ -1,18 +1,69 @@
...
+
+The ``mlflow`` client reads auth URI and token from environment variables.
```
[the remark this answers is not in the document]
alexjuda: Not urgent, but do you know which grammar applies here, making the definite article required?

docs/guides/mlflow.rst
alexjuda: Went with a shorter version, wdyt?

setup.cfg
alexjuda: OH YES

docs/examples/tests/test_mlflow_utilities.py
```diff
@@ -6,48 +6,255 @@
...
 class TestSnippets:
-    @staticmethod
-    def test_artifact_dir_local(monkeypatch: MonkeyPatch, tmp_path):
-        monkeypatch.delenv("ORQ_MLFLOW_ARTIFACTS_DIR", raising=False)
-        monkeypatch.setattr(
-            sdk.mlflow._connection_utils, "DEFAULT_TEMP_ARTIFACTS_DIR", tmp_path
-        )
+    class TestTrackingURIToken:
+        @staticmethod
+        def test_local(mocked_environ, monkeypatch):
```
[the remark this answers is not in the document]
alexjuda: Thanks, fixed in aed11b2457df6a81f295cd66ae8cae86dc02c4c7.

One unfortunate consequence of having the snippet tests outside of `tests` dir is we cannot easily share fixtures between `tests/` and `docs/` :/ I fiddled around with it but couldn't find any quick fixes. I guess we'll have to kick the can down the road one more time.

setup.cfg
alexjuda: b00477f

alexjuda:
Marking as draft because the tests need more fixing than I anticipated.

BenjaminMummery (approved):
LGTM

setup.cfg
BenjaminMummery: Nice :D

docs/guides/mlflow.rst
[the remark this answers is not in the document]
BenjaminMummery: The distinction here is between talking about a member of a category, and a category itself. We need the article to clarify grammatically that we're talking about a specific auth URI, rather than the concept of auth URIs. 

[omitted for brevity: a long digression about Plato and the world of forms]

There's a significant difference between the way a name operates, and the way a description does. The sentence would be fine if it was "a and b" or "Carlos and Simon" because these are labels that uniquely identify entities of a known type, variables in the first case and people in the second. They convey no information about what classes of thing a, b, Carlos and Simon belong to, the reader's knowledge and the context take care of that so all the name does is uniquely identify the subject - the reader already knows how a person is quantised and that Carlos is a person so we don't need the article. The reader's brain effectively reads it as "Carlos the person".

Here, however, "auth URI" has to tell you _what_ it is at the same time as telling you, effectively, _who_ it is. As such, we need the article to tell the reader that we're talking about a member of the category "auth URI" rather than the category itself. In the absence of the article, the default reading is that we're talking about an indeterminate quantity of _stuff_ that belongs to the category. As an example, consider the difference between "There is a deer on the tracks" and "There is deer on the tracks". This is why it's fine to omit or include the article for "environment variables" - for the purpose of this sentence, the abstract class of "environment variables" and the concrete members "the environment variables" are both valid.

The choice of article is less concrete, we use the indefinite article here and still be grammatically accurate. However the definite article is the appropriate choice in context - the client is reading a specific token whose identity is known in advance.

PR #277, opened 2023-08-04
imp: expose `dates` in the public API

alexjuda:
# The problem

`orquestra.sdk.schema.workflow_run.WorkflowRun` contained fields of type `Instant`. If a user manipulates `WorkflowRun` objects, they're likely to get into typecheck errors, because they need to work on `Instant` objects. However, `Instant` wasn't exposed in the public API, so the user would either have to:
1. Ignore the type checker error.
2. Import `Instant` from a private module, risking unwarranted breakage.

# This PR's solution

Exposes `dates` in the public API.

# Checklist

### DIFF
--- a/src/orquestra/sdk/dates/__init__.py
+++ b/src/orquestra/sdk/dates/__init__.py
@@ -1,0 +1,30 @@
+################################################################################
+# © Copyright 2023 Zapata Computing Inc.
+################################################################################
+"""
+Single place to handle datetimes and timezones without shooting yourself in the
+foot.
+"""
+from ._dates import (
+    Instant,
+    from_comps,
+    from_isoformat,
+    from_unix_time,
+    isoformat,
+    local_isoformat,
+    now,
+    unix_time,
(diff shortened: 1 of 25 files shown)


src/orquestra/sdk/dates/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+    "Instant",
+    "from_comps",
+    "from_isoformat",
+    "from_unix_time",
+    "isoformat",
+    "local_isoformat",
+    "now",
+    "unix_time",
+    "utc_from_comps",
+]
```
jamesclark-Zapata, line 30: We haven't thought about this API, we've been good at iterating on APIs before making public.
Can we make note in the changelog that it is subject to changes? and then for us to get together to figure out what's a good API for this.

alexjuda: Yeah, good point. I'll close this one and make a separate plan.

PR #282, opened 2023-08-08
Feat: Dry_run of a workflows

SebastianMorawiec:
# The problem
It would be useful to test the “decorators” of the tasks/workflows.

This could be done by running the workflow and each task is a no-op, but the resources, imports, etc. are all allocated/installed.
(https://zapatacomputing.atlassian.net/browse/ORQSDK-919)

# This PR's solution
Add new option to workflow runs that allows to run workflow without actually running the task code

# Checklist


src/orquestra/sdk/_base/_driver/_models.py
```diff
@@ -271,6 +271,7 @@ class CreateWorkflowRunRequest(pydantic.BaseModel):
 
     workflowDefinitionID: WorkflowDefID
     resources: Resources
+    dry_run: bool
```
alexjuda, line 274: **issue**: Is it supported on the Platform already? This field isn't in the [CE's API spec](https://github.com/zapatacomputing/workflow-driver/blob/5d030e5a71be088c86590ce63180e256b55b3c8e/openapi/src/schemas/CreateWorkflowRunRequest.yaml). Even if we want to add client code before the Platform is ready, the spec should be updated first.

alexjuda (changes requested):

src/orquestra/sdk/_base/_driver/_models.py
SebastianMorawiec: Its not supported - my intention is to wait for workflow-driver to support this feature before I merge this.

src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -152,7 +162,12 @@ def __call__(self, *wrapped_args, **wrapped_kwargs):
         for name, kwarg in wrapped_kwargs.items():
             kwargs[name] = self._unpack_argument(kwarg, name)
 
-        return self._user_fn(*args, **kwargs)
+        # if the dry_run mode is enabled, we don't execute user function, but just
+        # empty nop function returning Nones
+        if self._dry_run:
+            return self._nop_function_factory()()
+        else:
+            return self._user_fn(*args, **kwargs)
```
alexjuda, line 170: **nitpick**: this class (`ArgumentUnwrapper`) is about unwrapping arguments to make it suitable for interfacing with Ray tasks. The `dry_run` is an unrelated feature. Can you extract it to a different component? Functions are composable.

tests/runtime/ray/test_integration.py
```diff
@@ -112,11 +112,34 @@ def test_sets_run_id_from_env(
...
+            run_id = runtime.create_workflow_run(wf_def, None, False)
 
             runtime.stop_workflow_run(run_id)
             assert run_id == global_wf_run_id
 
+        def test_simple_workflow_dry_run(
+            self, monkeypatch: pytest.MonkeyPatch, runtime: _dag.RayRuntime
+        ):
+            wf_def = _example_wfs.exception_wf_with_multiple_values.model
+            run_id = runtime.create_workflow_run(wf_def, None, True)
```
alexjuda, line 124: **nitpick**: please use named arguments when passing literals.

tests/sdk/api/test_wf_run.py
```diff
@@ -1366,7 +1370,7 @@ def test_run(
         monkeypatch.setattr(_api._config.RuntimeConfig, "name", "auto")
         with raises:
             wf_def.run("in_process", workspace_id=workspace_id, project_id=project_id)
-            workflow_create_mock.assert_called_once_with(wf_def.model, expected)
+            workflow_create_mock.assert_called_once_with(wf_def.model, expected, False)
```
alexjuda, line 1373: nitpick: please use named arguments when passing literals.

tests/sdk/driver/test_ce_runtime.py
```diff
@@ -82,7 +82,7 @@ def test_happy_path(
         mocked_client.create_workflow_run.return_value = workflow_run_id
 
         # When
-        wf_run_id = runtime.create_workflow_run(my_workflow.model, None)
+        wf_run_id = runtime.create_workflow_run(my_workflow.model, None, False)
```
alexjuda, line 85: nitpick: please use named arguments when passing literals.

tests/sdk/driver/test_client.py
```diff
@@ -1040,7 +1040,7 @@ def test_invalid_request(
                 )
 
                 with pytest.raises(_exceptions.InvalidWorkflowRunRequest):
-                    _ = client.create_workflow_run(workflow_def_id, resources)
+                    _ = client.create_workflow_run(workflow_def_id, resources, False)
```
alexjuda, line 1043: nitpick: please use named arguments when passing literals.

tests/sdk/test_in_process_runtime.py
```diff
@@ -77,14 +77,16 @@ def test_secret_inside_ir(
     get_secret.return_value = mocked_secret
     monkeypatch.setattr(_client.SecretsClient, "get_secret", get_secret)
 
-    run_id = runtime.create_workflow_run(wf_with_secrets().model, None)
+    run_id = runtime.create_workflow_run(wf_with_secrets().model, None, False)
```
alexjuda, line 80: nitpick: please use named arguments when passing literals.

tests/runtime/ray/test_build_workflow.py
```diff
@@ -391,7 +399,7 @@ def test_positional(
         ):
             fn = Mock()
             arg_unwrapper = _build_workflow.ArgumentUnwrapper(
-                fn, args_artifact_nodes, {}, True
+                fn, args_artifact_nodes, {}, True, 1, False
```
alexjuda, line 402: nitpick: please use named arguments when passing literals.

alexjuda (changes requested):

tests/sdk/api/test_wf_run.py
SebastianMorawiec: For all the comments about passing literals: I can do changes with function call, cant apply them to mock.asserts - if parameter is passed as arg, it has to be checked as arg, not kwarg

src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -142,6 +146,12 @@ def _unpack_argument(
         else:
             assert_never(arg)
 
+    def _nop_function_factory(self):
+        def nop():
+            return (None,) * self._n_outputs
```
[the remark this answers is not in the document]
SebastianMorawiec: Ive also checked your example of:
```
@sdk.task
def foo():
    return "abc"
```

- Right now we treat strings as non-subscribtable. So return "Abc" is the same for us as return <any literal like None>

src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -426,6 +453,8 @@ def make_ray_dag(
             n_outputs=_compat.n_outputs(task_def=user_task, task_inv=invocation),
             project_dir=project_dir,
             user_fn_ref=user_task.fn_ref,
+            output_metadata=user_task.output_metadata,
```
alexjuda, line 456: **nitpick**: shouldn't `n_outputs` be replaced with this parameter?

alexjuda (approved):

PR #283, opened 2023-08-08
int: prototype makefile cleanup

alexjuda:
# The problem

Our nested makefile with z-quantum-actions subtree is outdated and needs an overhaul. The original idea [described on Jira](https://zapatacomputing.atlassian.net/browse/ORQSDK-456) was to replace the subtree with a makefile fetched from another source.

Meanwhile, it doesn't make a lot of sense either. The makefile in this repo overrides all targets anyway (the only exception: `make clean`).

# This PR's solution

POC of an alternative solution, other than the one originally specified in the ticket.

* Removes the original makefile inheritance.
* Adds missing make variables.

# Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,17 @@
 ################################################################################
-# © Copyright 2021-2022 Zapata Computing Inc.
+# © Copyright 2021-2023 Zapata Computing Inc.
 ################################################################################
-include subtrees/z_quantum_actions/Makefile
 
 # We need to override test commands in 'make test' and 'make coverage' to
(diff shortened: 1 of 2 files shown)


max-radin (changes requested):
I think we want to avoid modifying `subtrees/z_quantum_actions/Makefile` because then the subtree will be out of sync with the `z_quantum_actions` repos. Otherwise looks good to me!

alexjuda:
The `subtree` directory is deleted in another PR, https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/285.

max-radin (approved):

PR #290, opened 2023-08-13
internal: Print coverage only on matrix jobs with print_coverage set

jamesclark-Zapata:
# The problem

Our code coverage action could post twice, due to a race condition when two jobs completed at the same time.

# This PR's solution

Added an extra item to the matrix that means coverage printing only runs on the `{ os: "ubuntu-latest", python: "3.9" }` job.

# Checklist


SebastianMorawiec (approved):

SebastianMorawiec:
Suggestion: Could you add a comment why are we adding this condition?

PR #293, opened 2023-08-16
feat: Task logs from CE

jamesclark-Zapata:
# The problem

We need to expose task logs from CE to our users.

# This PR's solution

- Implement the task logs endpoint.
- Update workflow logs endpoint to use task logs

# Checklist


src/orquestra/sdk/_base/_driver/_ce_runtime.py
```diff
@@ -525,17 +535,58 @@ def get_workflow_logs(self, wf_run_id: WorkflowRunId) -> WorkflowLogs:
         for sys_m in sys_messages:
             system_logs.add_line_by_stream(LogStreamType.STDOUT, str(sys_m.log))
 
+        # If we don't have any logs and we have ray worker logs, we should use these
+        # as the "unknown" task invocation logs. This is to ensure backwards compat
```
alexjuda, line 539: **nitpick**: `"UNKNOWN TASK INV ID"` was a temporary placeholder I'd prefer to get rid of as soon as we support task log correlation. We already have `.other` logs as a "plan B" for unforseen situations. Is there a way to use `.other` in this scenario?

src/orquestra/sdk/_base/_driver/_ce_runtime.py
```diff
@@ -525,17 +535,58 @@ def get_workflow_logs(self, wf_run_id: WorkflowRunId) -> WorkflowLogs:
...
+            raise exceptions.InvalidWorkflowRunLogsError(
+                f"Failed to decode logs for workflow run with id `{wf_run_id}` "
+                f"and task invocation ID `{task_inv_id}`. "
+                "Please report this as a bug."
+            ) from e
+
+        task_logs = LogAccumulator()
+        for m in messages:
+            path = Path(m.log_filename or "log.out")
+            stream = LogStreamType.by_file(path)
```
alexjuda, line 587: **non-blocking**: WDYT? `"log.out"` is a magic constant but we don't need `path` for anything else here.
```suggestion
            stream = LogStreamType.by_file(Path(m.log_filename)) if m.log_filename else LogStreamType.STDOUT
```

src/orquestra/sdk/_base/_driver/_exceptions.py
```diff
@@ -149,14 +149,26 @@ def __init__(self, workflow_run_id: str):
         super().__init__(workflow_run_id)
 
 
+class TaskRunLogsNotFound(Exception):
+    """
+    Raised when a Task Run's Logs cannot be found
+    """
+
+    def __init__(self, workflow_run_id: str, task_invocation_id: str):
```
alexjuda, line 157: **non-blocking,question**: is there a reason we're not using `WorkflowRunId` and `TaskInvocationId` as type annotations in this module?

src/orquestra/sdk/_base/_driver/_models.py
```diff
@@ -438,7 +440,10 @@ class Message(pydantic.BaseModel):
     Single line content.
     """
 
-    ray_filename: RayFilename
+    # Pydantic v2 lets us do multiple aliases
+    # TODO: ORQSDK-956 - update Pydantic v2 and use AliasChoice
+    log_filename: Optional[LogFilename] = None
+    ray_filename: Optional[RayFilename] = None
```
alexjuda, line 446: **issue**: `ray_filename` is a required property in the [API schema](https://github.com/zapatacomputing/workflow-driver/blob/17a53d42f42f1f90ee1c1c1eba2032117a5380e5/openapi/src/resources/workflow-run-logs.yaml#L28). It's a good practice to catch invalid data as soon as it enters our subsystem. Marking the property as optional would prevent us from identify the error.

src/orquestra/sdk/exceptions.py
```diff
@@ -156,6 +156,12 @@ class InvalidWorkflowRunLogsError(BaseRuntimeError):
     pass
 
 
+class TaskRunLogsNotFound(NotFoundError):
+    """Raised when a task run logs cannot be found, or the ID is invalid"""
+
```
alexjuda, line 161: **non-blocking,suggestion**: WDYT about adding `wf_run_id` and `task_inv_id` as fields here?

src/orquestra/sdk/_base/_driver/_models.py
```diff
@@ -438,7 +440,10 @@ class Message(pydantic.BaseModel):
     Single line content.
     """
 
-    ray_filename: RayFilename
+    # Pydantic v2 lets us do multiple aliases
+    # TODO: ORQSDK-956 - update Pydantic v2 and use AliasChoice
+    log_filename: Optional[LogFilename] = None
```
alexjuda, line 445: **issue**: there's no `log_filename` in the schema for [TaskRunLogs](https://github.com/zapatacomputing/workflow-driver/blob/17a53d42f42f1f90ee1c1c1eba2032117a5380e5/openapi/src/resources/task-run-logs.yaml#L28) nor [WorkflowRunLogs](https://github.com/zapatacomputing/workflow-driver/blob/17a53d42f42f1f90ee1c1c1eba2032117a5380e5/openapi/src/resources/workflow-run-logs.yaml#L26) entries.

src/orquestra/sdk/_base/_driver/_models.py
```diff
@@ -438,7 +440,10 @@ class Message(pydantic.BaseModel):
     Single line content.
     """
 
-    ray_filename: RayFilename
+    # Pydantic v2 lets us do multiple aliases
```
alexjuda, line 443: **thought**: so far, `Message` was only about WorkflowRunLogs entries (IIRC). Now, we're starting to use it for TaskRunLogs entries, which look similar as of today, but it might get messy when two endpoint schemas diverge.

alexjuda (changes requested):
Blocking to make sure our specs are up to date

src/orquestra/sdk/exceptions.py
jamesclark-Zapata: will do

src/orquestra/sdk/_base/_driver/_ce_runtime.py
jamesclark-Zapata: Should we always put in these worker logs? they're basically empty after we intercept stdout/stderr

src/orquestra/sdk/_base/_driver/_models.py
jamesclark-Zapata: TaskRunLogs are out of date, Emre added this last night.

src/orquestra/sdk/_base/_driver/_ce_runtime.py
alexjuda: > Should we always put in these worker logs

Not sure if I understand correctly. If you mean taking worker logs from the API response and exposing this data under `WorkflowLogs.other` then yeah, sounds good. Otherwise, please elaborate :)

jamesclark-Zapata: What I meant was should we always do this or should we only add the worker logs when we didn't have any task run logs?

alexjuda (approved):
Please remember to link the updated OpenAPI spec!

PR #299, opened 2023-08-18
Feat!: ORQ_CURRENT_CONFIG as default config when using "auto" 

SebastianMorawiec:
# The problem

Task authors like the QML team have to hardcode the config name in the task decorator, if they’re using GithubImport with private repos.
https://zapatacomputing.atlassian.net/browse/ORQSDK-942

# This PR's solution

Allow to set ORQ_CURRENT_CONFIG env variable which will be used locally when "auto" config is used.

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_api/_wf_run.py
+++ b/src/orquestra/sdk/_base/_api/_wf_run.py
@@ -138,6 +138,7 @@ class WorkflowRun:
         config: t.Union[RuntimeConfig, str],
         workspace_id: t.Optional[WorkspaceId] = None,
         project_id: t.Optional[ProjectId] = None,
+        project_dir: t.Optional[t.Union[str, Path]] = None,
     ):
         """
         Start workflow run from its IR representation
@@ -149,11 +150,12 @@ class WorkflowRun:
                 the name of a saved configuration.
             workspace_id: ID of the workspace for workflow - supported only on CE
             project_id: ID of the project for workflow - supported only on CE
-
+            project_dir: the path to the project directory. If omitted, the current
+                working directory is used.
         """
         _config = resolve_config(config)
 
-        runtime = _config._get_runtime()
+        runtime = _config._get_runtime(project_dir)
 
         assert runtime is not None
 
--- a/src/orquestra/sdk/_base/_config.py
+++ b/src/orquestra/sdk/_base/_config.py
@@ -241,11 +241,14 @@ def _resolve_auto_config(config_name) -> RuntimeConfiguration:
     # On studio we short-circuit to use internal URIs
     if is_passport_file_available():
         return _resolve_remote_auto_config(config_name)
-
-    if CURRENT_CONFIG_ENV in os.environ:
+    elif CURRENT_CONFIG_ENV in os.environ:
         return _resolve_local_auto_config(os.environ[CURRENT_CONFIG_ENV])
-
-    return LOCAL_RUNTIME_CONFIGURATION
+    else:
+        # we are not in the cluster, and default config env was not set. Error out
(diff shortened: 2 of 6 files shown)


src/orquestra/sdk/_base/_api/_wf_run.py
```diff
@@ -138,6 +138,7 @@ def start_from_ir(
         config: t.Union[RuntimeConfig, str],
         workspace_id: t.Optional[WorkspaceId] = None,
         project_id: t.Optional[ProjectId] = None,
+        project_dir: t.Optional[t.Union[str, Path]] = None,
```
alexjuda, line 141: **question**: what's the reason for `project_dir`? We used this in the past, but it's a *distant past* and I don't know any use case for anymore. Let's avoid cargo culting if we can.

alexjuda (changes requested):

src/orquestra/sdk/_base/_api/_wf_run.py
SebastianMorawiec: The only reason Ive added project_dir to this function, is so the code of this function can be reused in the WorkflowDef.run() without worrying about any backward incompatibility.
I do agree project_dir is a concept of a past - we might want to add a ticket to investigate if we can remove it, but here it is added just to reduce code duplication between WorkflowDef.run and WorkflowRun.start_from_ir()

src/orquestra/sdk/_base/_config.py
```diff
@@ -214,9 +215,39 @@ def _resolve_auto_config(config_name) -> RuntimeConfiguration:
...
+
+def _resolve_auto_config(config_name) -> RuntimeConfiguration:
+    # On studio we short-circuit to use internal URIs
+    if is_passport_file_available():
+        return _resolve_remote_auto_config(config_name)
+
+    if CURRENT_CONFIG_ENV in os.environ:
+        return _resolve_local_auto_config(os.environ[CURRENT_CONFIG_ENV])
+
+    return LOCAL_RUNTIME_CONFIGURATION
```
jamesclark-Zapata, line 248: **suggestion:** I think we should discuss this fallback. I think we should break previous behaviour and have "auto" fail when `ORQ_CURRENT_CONFIG` is not set.
This is because it is likely that someone will forget to set the environment variable and be confused why their workflow isn't running on cluster X. However, it's running locally.

jamesclark-Zapata (changes requested):

src/orquestra/sdk/_base/_config.py
```diff
@@ -214,9 +215,42 @@ def _resolve_auto_config(config_name) -> RuntimeConfiguration:
...
+    if is_passport_file_available():
+        return _resolve_remote_auto_config(config_name)
+    elif CURRENT_CONFIG_ENV in os.environ:
+        return _resolve_local_auto_config(os.environ[CURRENT_CONFIG_ENV])
+    else:
+        # we are not in the cluster, and default config env was not set. Error out
+        raise exceptions.RuntimeConfigError(
+            f"Config {config_name} only works on the studio or when the "
+            f"{CURRENT_CONFIG_ENV} env variable is set."
+        )
```
jamesclark-Zapata, line 251: **suggestion:**
```suggestion
        raise exceptions.RuntimeConfigError(
            f"Using '{config_name}' as the config name requires that "
            f"you're  using Studio or that the '{CURRENT_CONFIG_ENV}' "
            "environment variable is set.\n"
            "For example, if you want to use a local Ray cluster, "
            f"set `{CURRENT_CONFIG_ENV}=local`."
        )
```

jamesclark-Zapata (approved):

alexjuda (approved):

PR #324, opened 2023-10-03
feat!:  add orquestra sdk head node current version as a requirement in ray runtime environments

BenjaminMummery:
# The problem

When a workflow runs on CE/Ray, there is a head node and worker nodes. The head node will use the version of orquestra-sdk in the environment that launched Ray (or `orq up`). Each worker can install a new python environment per task.

It’s possible for the dependencies of a task to install the wrong version of orquestra-sdk inside a worker, e.g. by a wrong custom image or via poorly defined PythonImports.

The outcome of ORQSDK-848 was to add orquestra-sdk==<head node version>to the runtime environment of all tasks.

If a different version is brought in via the user’s imports, pip will complain and will quit. See the ADR in 848 for more details.

# This PR's solution

Adds the currently installed[^1] SDK version to the runtime environment pip requirements.

[^1]: In order to avoid trying to install `orquestra-sdk==0.57.1.dev3+g3ef9f57.d20231003`, we strip out prerelease versioning (in this example, the correct dependency `orquestra-sdk==0.57.0` would be added instead). This creates an issue in our CI where the installed SDK version detected by setuptools is `0.1-dev...`. I've therefore implemented an escape hatch where rather than the semantically nonsensical `orquestra-sdk==0.0`, this case adds `orquestra-sdk` without a version. This somewhat compromises the intent of this PR, but I can't think of a scenario outside our CI where this scenario would occur.

As a result of this change, specifying the SDK in the task dependencies is now meaningless - it will always be installed regardless of what you do. Therefore, before adding the current version to the list we check for a dependency on the SDK and, if it's there, remove it from the list and warn the user that it has been overruled. This means that rather than getting a pip error when listing the SDK as a requirement, the user gets an informative message and the workflow runs as intended. This is intended to reduce the confusion of pip telling them that there are multiple clashing requirements when they've only specified one.

I've also updated the `dependency installation` guide so it no longer indicates that people can specify sdk versions in their dependencies.

# Supplemental

I've also had to increase the polling timeout for integration tests - all tasks now require _at minimum_ the SDK to install, so they take a little longer.

# Checklist

[ORQSDK-937]

[ORQSDK-937]: https://zapatacomputing.atlassian.net/browse/ORQSDK-937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 🚨 *Breaking Changes*
 
+* The `orquestra-sdk` version used to submit a workflow is automatically added as a dependency for task execution environments. Consequently, specifying the SDK as a dependency in the `sdk.task()` decorator is no longer supported, and such declarations will be ignored.
+
 🔥 *Features*
 
 🧟 *Deprecations*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 🚨 *Breaking Changes*
 
-* The `orquestra-sdk` version used to submit a workflow is automatically added as a dependency for task execution environments. Consequently, specifying the SDK as a dependency in the `sdk.task()` decorator is no longer supported, and such declarations will be ignored.
+* The `orquestra-sdk` version used to submit a workflow is automatically added as a dependency for task execution environments. Specifying the SDK as a dependency in the `sdk.task()` decorator will be ignored.
 
 🔥 *Features*
 
--- a/src/orquestra/sdk/_ray/_build_workflow.py
+++ b/src/orquestra/sdk/_ray/_build_workflow.py
@@ -8,11 +8,12 @@ import time
 import typing as t
 import warnings
 from functools import singledispatch
-from orquestra.sdk.packaging import get_installed_version
 from pathlib import Path
 
 import pydantic
 from typing_extensions import assert_never
+
+from orquestra.sdk.packaging import get_installed_version
 
 from .. import exceptions, secrets
 from .._base import _exec_ctx, _git_url_utils, _graphs, _services, dispatch, serde
--- a/src/orquestra/sdk/_ray/_build_workflow.py
+++ b/src/orquestra/sdk/_ray/_build_workflow.py
@@ -22,6 +22,7 @@ from .._base._env import (
     RAY_SET_CUSTOM_IMAGE_RESOURCES_ENV,
 )
 from .._base._logs import _markers
+from .._base._regex import SEMVER_REGEX
 from ..kubernetes.quantity import parse_quantity
 from ..schema import ir, responses, workflow_run
 from . import _client, _id_gen
@@ -360,8 +361,10 @@ def _import_pip_env(
             *(task_def.dependency_import_ids or []),
         )
     ]
-    sdk_version = get_installed_version("orquestra-sdk")
 
+    current_sdk_version: str = _normalise_prerelease_version(
+        get_installed_version("orquestra-sdk")
+    )
     sdk_dependency = None
     pip_list = [
         chunk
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 🚨 *Breaking Changes*
 
+* The `orquestra-sdk` version used to submit a workflow is automatically added as a dependency for task execution environments. Consequently, specifying the SDK as a dependency in the `sdk.task()` decorator is no longer supported, and such declarations will be ignored.
```
jamesclark-Zapata, line 7: **suggestion:** ‏This was never supported, I'd rephrase this.

src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -3,9 +3,12 @@
 ################################################################################
 """Translates IR workflow def into a Ray workflow."""
 import os
+import re
 import time
 import typing as t
+import warnings
 from functools import singledispatch
+from importlib.metadata import version
```
jamesclark-Zapata, line 11: **suggestion:** ‏We already use expose this https://github.com/zapatacomputing/orquestra-workflow-sdk/blob/main/src/orquestra/sdk/packaging/__init__.py#L16
```suggestion
from orquestra.sdk.packaging import get_installed_version
```

CHANGELOG.md
BenjaminMummery: How about:
```suggestion
* The `orquestra-sdk` version used to submit a workflow is automatically added as a dependency for task execution environments. Specifying the SDK as a dependency in the `sdk.task()` decorator will be ignored.
```

tests/runtime/ray/test_build_workflow.py
```diff
@@ -412,3 +413,129 @@ def test_kwargs(
...
+    def test_with_no_dependencies():
+        # Given
+        wf = make_workflow_with_dependencies([]).model
+        task_inv = [inv for inv in iter_invocations_topologically(wf)][0]
+
+        # When
+        pip = _build_workflow._import_pip_env(task_inv, wf)
+
+        # Then
+        assert pip == [f"orquestra-sdk=={version('orquestra-sdk')}"]
```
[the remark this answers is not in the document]
BenjaminMummery: we could indeed - [efb0498](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/324/commits/efb0498aadf25c1ff817179da83b87712c6f2646)

src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -343,7 +359,28 @@ def _import_pip_env(ir_invocation: ir.TaskInvocation, wf: ir.WorkflowDef):
...
+    # If the task definition includes the SDK, warn the user that this does nothing.
+    if sdk_dependency:
+        warnings.warn(
+            f"The definition for task `{ir_invocation.task_id}` "
+            f"declares `{sdk_dependency[0]}` as a dependency. "
+            f"The current SDK version ({sdk_version}) is automatically installed in "
+            "task environments. "
+            "The specified dependency will be ignored.",
+            DeprecationWarning,
+        )
```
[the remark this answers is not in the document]
BenjaminMummery: w.r.t. the warn / fail choice - I have no strong opinion either way. I went with this option because it shortens the time between the user and their results.
w.r.t. transitive dependencies - yep, this is entirely opportunistic. It's easy to give the user an informative warning here so we might as well.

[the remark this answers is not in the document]
jamesclark-Zapata: I'm happy to go with warn for now, it's an easy change to make it a failure.
Can you change the warning class to a custom warning class? Something like `OrquestraSDKVersionMismatchWarning` or a better name!

src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -343,7 +361,52 @@ def _import_pip_env(ir_invocation: ir.TaskInvocation, wf: ir.WorkflowDef):
             *(task_def.dependency_import_ids or []),
         )
     ]
-    return [chunk for imp in imports for chunk in _pip_string(imp)]
+
+    current_sdk_version: t.Optional[str] = _normalise_prerelease_version(
+        get_installed_version("orquestra-sdk")
+    )
```
jamesclark-Zapata, line 367: I'm not sure about this.
I think we should just ignore the dependency and not add it to the `pip_list` if it looks like a development version.
If we install a previous version, it's possible that you don't end up testing the feature you just developed.

BenjaminMummery: Good idea, [5631ffa](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/324/commits/5631ffab578770be00bd9d31838fa81b72a3ad8b) makes it so that dev versions of the sdk don't add a dependency to the tasks.

src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -385,9 +385,12 @@ def _import_pip_env(
             exceptions.OrquestraSDKVersionMismatchWarning,
         )
 
-    return pip_list + [
-        "orquestra-sdk" + (f"=={current_sdk_version}" if current_sdk_version else "")
-    ]
+    # Don't add sdk dependency if submitting from a prerelease or dev version.
+    parsed_sdk_version = version.parse(current_sdk_version)
```
jamesclark-Zapata, line 389: **thought:** ‏We have something else that wraps this, but it is not suitable. I think this is OK for now.

jamesclark-Zapata (approved):

PR #339, opened 2023-11-16
Imp: Remove usage of project_dir arguments in SDK

SebastianMorawiec:
# The problem
https://zapatacomputing.atlassian.net/browse/ORQSDK-971
Project_dir argument was a leftover from VS studio times. Now it is obsolete and makes the unnecessary complex.

# This PR's solution

Remove project_dir internally. Deprecate it on user-API functions

# Checklist

### DIFF
--- a/src/orquestra/sdk/_ray/_build_workflow.py
+++ b/src/orquestra/sdk/_ray/_build_workflow.py
@@ -195,7 +195,6 @@ def _make_ray_dag_node(
     ray_kwargs: t.Mapping[str, t.Any],
     args_artifact_nodes: t.Mapping,
     kwargs_artifact_nodes: t.Mapping,
-    project_dir: t.Optional[Path],
     user_fn_ref: t.Optional[ir.FunctionRef],
     output_metadata: t.Optional[ir.TaskOutputMetadata],
     dry_run: bool,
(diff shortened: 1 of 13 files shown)


src/orquestra/sdk/_ray/_build_workflow.py
```diff
@@ -213,14 +212,14 @@ def _make_ray_dag_node(
...
             see ArgumentUnwrapper.
-        project_dir: the working directory the workflow was submitted from.
         user_fn_ref: function reference for a function to be executed by Ray.
             if None - executes data aggregation step.
         output_metadata: output metadata for the user task function.
             Keeps number of outputs and if the output is subscriptable.
         dry_run: Run the task without actually executing any user code.
             Useful for testing infrastructure, dependency imports, etc.
     """
+    current_path = Path.cwd()
```
alexjuda, line 222: **thought**: it looks we're safe for now, but we'll have to be careful about where this code is being executed. If it was wrapped in some Ray remote, `Path.cwd()` is likely to be resolved to PWD at the time of running `orq up` which might be different from where the project files need to be loaded

alexjuda: **thought 2**: maybe some time in the future we can get of the `sys.path` manipulation in the first place?

alexjuda (approved):
So good!

src/orquestra/sdk/_ray/_build_workflow.py
SebastianMorawiec: That would be great, but Im honestly not sure we are even able to do that due to how we rely on users code to be on the path to launch scripts and other stuff

PR #340, opened 2023-11-16
Int: Remove optional str from runtime name

SebastianMorawiec:
# The problem

https://zapatacomputing.atlassian.net/browse/ORQSDK-996
Name for config was optional for historical reasons. Now we always pass the name in our factory methods.

# This PR's solution

Initial idea was to deprecate usage of `name = None` BUT
I found out that our `RuntimeConfig` has convention of factory methods, and using `__init__` should not be considered as public API due to `bypass_factory_methods` setting (plus there is bunch of comments saying to not use it)
So as a solution ive already changed the `name` parameter from `optional[str]` to `str`

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/_api/_config.py
+++ b/src/orquestra/sdk/_base/_api/_config.py
@@ -57,7 +57,7 @@ class RuntimeConfig:
                 "`RuntimeConfig.ce()` for Compute Engine. \n"
             )
 
-        self._name: str = name
+        self._name = name
         try:
             self._runtime_name: RuntimeName = RuntimeName(runtime_name)
         except ValueError as e:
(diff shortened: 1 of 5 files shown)


src/orquestra/sdk/_base/_api/_config.py
```diff
@@ -58,7 +58,7 @@ def __init__(
                 "`RuntimeConfig.ce()` for Compute Engine. \n"
             )
 
-        self._name = name
+        self._name: str = name
```
alexjuda, line 61: **question**: Is this type hint needed? I'd think type checkers infer the types of attributes from the init's argument types

alexjuda (approved):

PR #342, opened 2023-11-24
imp: resolve import pip strings once instead of per import

jamesclark-Zapata:
# The problem

- We create pip strings for all imports an invocation uses, for every invocation.
- Even if all invocations use the same import, we have to build the pip string.
- For Git imports that use a secret, this meant calling the secrets API to get the secret every time.
- If you have numerous tasks that use the same Git import with a secret, this can slow down workflow submission.

# This PR's solution

Before iterating over the task invocations, we pre-compute the pip strings for all imports.
We then re-use these pip strings for all invocations.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 👩‍🔬 *Experimental*
 
 🐛 *Bug Fixes*
+* Secrets are only fetched once for imports in a workflow, instead of for each task invocation.
 
 * Workaround for Ray cluster not starting because of a missing dependency, `async_timeout`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,13 +14,13 @@
 👩‍🔬 *Experimental*
 
 🐛 *Bug Fixes*
-* Secrets are only fetched once for imports in a workflow, instead of for each task invocation.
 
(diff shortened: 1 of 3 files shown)


SebastianMorawiec (changes requested):
Question: There is bunch of changes in test files, but I don't see a test that checks this behaviour in particular. Unless I am missing something, please add testcase verifying that it actually fixes underlying issue (also to prevent regression)

Nitpick: Is this really a Fix? I'd rather say its improvement :)

jamesclark-Zapata:
> Question: There is bunch of changes in test files, but I don't see a test that checks this behaviour in particular. Unless I am missing something, please add testcase verifying that it actually fixes underlying issue (also to prevent regression)
Yeah, I ended up indenting a lot which makes the diff hard to read. This test is the new one:
https://github.com/zapatacomputing/orquestra-workflow-sdk/blob/8bcf12da20b5f2162e0a11662289aedf8e5ba8c2/tests/runtime/ray/test_build_workflow.py#L145-L165

> Nitpick: Is this really a Fix? I'd rather say its improvement :)
I'd say it was broken before, but I'll change it to an improvment.

SebastianMorawiec (approved):
👍

PR #347, opened 2023-12-15
Feat: New API `WorkflowRun.get_artifact()` and `WorkflowRun.get_artifact_serialized()` to get single task output

SebastianMorawiec:
# The problem
We lacked the ability to get single artifact based on invocation ID. This caused us to always fetch all the artifacts, which is problematic for workflows with lots of artifacts where each individual artifacts is memory-heavy 

# This PR's solution
Add new API to get single artifact - now only implemented on Ray Runtime

# Checklist

### DIFF
--- a/src/orquestra/sdk/_base/abc.py
+++ b/src/orquestra/sdk/_base/abc.py
@@ -117,6 +117,29 @@ class RuntimeInterface(ABC, LogReader):
         raise NotImplementedError()
 
     @abstractmethod
+    def get_output(
+        self, workflow_run_id: WorkflowRunId, task_invocation_id: TaskInvocationId
+    ) -> WorkflowResult:
+        """Returns single output for a workflow run.
+
+        This method returns single artifact for given task.
+        When the task failed or was not yet completed, this method throws an exception
+
+        Careful: This method does NOT return status of a task.
+        Verify it beforehand to make sure if task failed/succeeded/is running.
+        You might get an exception
+
+        Args:
+            workflow_run_id: ID identifying the workflow run.
+            task_invocation_id: ID identifying single task run
+
+        Returns:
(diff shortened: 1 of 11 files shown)


alexjuda (approved):

alexjuda:
Took me a while to consider possible edge cases related to tasks with multiple outputs (`n_outputs=2`) but your unit tests and the `RuntimeInterface` docstring seem to cover it

src/orquestra/sdk/_base/abc.py
```diff
@@ -116,6 +116,29 @@ def get_available_outputs(
...
+        Verify it beforehand to make sure if task failed/succeeded/is running.
+        You might get an exception
+
+        Args:
+            workflow_run_id: ID identifying the workflow run.
+            task_invocation_id: ID identifying single task run
+
+        Returns:
+            Whatever the task function returned, independent of the
+            ``@task(n_outputs=...)`` value.
```
alexjuda, line 138: **suggestion**: Just for peace of mind, could you add an integration test that verifies that our Ray runtime implementation functions correctly with a task that has `n_outputs` > 1?

alexjuda: By "function correctly" I mean it should be consistent with this docstring.
Given:
* a task def with `n_outputs=2`
* task returns a `"foo", "bar"`
Expect:
* `runtime.get_output(...)[0] == "foo"`
* `runtime.get_output(...)[1] == "bar"`

SebastianMorawiec: rutnime.get_output[...] woudlnt work, as it returns JSON-encoded thing rather than tuple. But Ive added test to verify everythign works correctly

PR #352, opened 2023-12-19
int: use pyright as PR requirement

alexjuda:
# The problem

* There type errors that `mypy` doesn't find, but `pyright` can.
* We're running `pyright` on the CI already, but we don't require it, so the list of type errors naturally goes up.
* (What triggered me doing this): sometimes, a 3rd-party library doesn't expose its interface/typing/type stubs correctly. Example: `pyarrow` ([log](https://github.com/zapatacomputing/orquestra-workflow-sdk/actions/runs/7261022717/job/19781366657?pr=351)). This makes `mypy` return errors, but `pyright` doesn't complain.
* `mypy` is much slower to run than `pyright` (1min 30s vs 19s for the full `make style`). This is especially painful in local development.

# This PR's solution

* Uses `pyright` as the required type checker on CI.
* Adds a list of files that currently contain type errors. This allows us to split the type fixing work across weeks, while preventing new type errors to occur in the not-excluded files.

See also https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/353 for an example PR that narrows down the excludelist.

# Checklist


pyproject.toml
```diff
@@ -100,8 +100,63 @@ count = true
 arg-type-hints-in-docstring = false
 
 [tool.pyright]
-# pyright 1.1.340 and newer hangs up when analyzing this file.
 exclude = [
-    "./src/orquestra/sdk/_base/_traversal.py"
+    "src/orquestra/sdk/_base/_driver/_client.py",
```
jamesclark-Zapata, line 104: ooft. Excluding this many files may cause us problems if we never go back and fix the issues.

Did you enumerate the issues somewhere (like Slack or Confluence)?

alexjuda: Listed: https://zapatacomputing.atlassian.net/wiki/spaces/ORQSRUN/pages/842105015/Pyright+Type+Errors

alexjuda: At the moment we're ignoring pyright errors for all files. You can treat my PR as _enabling_ type checks for all the remaining files not listed in the excludelist.

SebastianMorawiec: IMHO I think this is amazing strategy to move forward. We are enabling check for "what works for now". And later during either by ticket-by-ticket or during innovation time or whatever we can start removing exceptions one by one basically.

SebastianMorawiec (changes requested):
I do not want to remove MYPY as exclude pyright list is so long. I agree with replacement but right now removing mypy as requirement would mean we do not have style checker at all. Just add pyright with exclude list, but leave mypy as is for now

alexjuda:
@SebastianMorawiec good point! [69ebb9](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/352/commits/69ebb9f6b03fbf8911366017c2b527db193e8eb8)

SebastianMorawiec (approved):
lgtm

PR #360, opened 2024-01-04
Int: Fix bunch of pyright issues

SebastianMorawiec:
# The problem

We had to add tons of files to filterlist for pyright to pass.

# This PR's solution

I went one by one on some of the files and fixed bunch of individual issues
Removed fixed files from exception list

# Checklist


src/orquestra/sdk/_base/_logs/_interfaces.py
```diff
@@ -12,8 +12,8 @@
 
 @dataclass(frozen=True)
 class LogOutput:
-    out: t.Sequence[str]
-    err: t.Sequence[str]
```
SebastianMorawiec, line 16: A comment on that change: Its good to have general interface, but we later in the code used + operator which is specifically defined for the lists.

alexjuda (approved):
🔥

src/orquestra/sdk/_base/cli/_arg_resolvers.py
```diff
@@ -438,7 +438,7 @@ def resolve_limit(
                     "Enter maximum number of results to display. "
                     "If 'None', all results will be displayed."
                 ),
-                default="None",
+                default=None,
```
alexjuda, line 441: I just saw it mentioned in the tests. IIRC something's wrong with the overall typing here, as "None" should be an accepted output (as much as I don't like this)

SebastianMorawiec: If I wasnt mistaken, the correct type here is Optional[int], so no idea how that would work with string type passed out tbh.
None type seems to be better anyway

PR #365, opened 2024-03-01
internal: Kill setup.cfg

jamesclark-Zapata:
# The problem

We use a strange hybrid of `setup.cfg` and `pyproject.toml` and miss standard fields in `pyproject.toml`, such as `name`.

# This PR's solution

This removes `setup.cfg` and uses the standard `pyproject.toml`.
Additionally, this nukes the pinned `setuptools` and uses a recent version of `hatchling` to build the package.

# Checklist


pyproject.toml
```diff
@@ -1,16 +1,139 @@
...
+    "wrapt",
+    # Configuration file locking
+    "filelock>=3.3.2",
+    # Parse version strings
+    "packaging>=21",
+    # For different import types in SDK
+    "GitPython",
+    "pip-api==0.0.30",
+    # Python Standard Library Backports
+    "importlib_metadata~=4.0",
```
alexjuda, line 45: **thought**: do you think we should also (in another PR) change all `~=` into `>=`?

alexjuda (approved):

pyproject.toml
jamesclark-Zapata: I think we should only do this if/when we have "nightly" runs of the test suite.

Otherwise, an update may be released for a dependent package with a breaking change, which bumps semver, and we miss it.

Perhaps we should do this already 😬

alexjuda: Nightly runs: great idea! I think it's just a matter of combining [cron triggers](https://jasonet.co/posts/scheduled-actions/) with disabling pip cache (or another approach that ensures we get the latest and greatest dependencies)

jamesclark-Zapata: ORQSDK-1022

PR #366, opened 2024-03-04
fix: Use a `-cuda` image when requesting a GPU

jamesclark-Zapata:
# The problem

Requesting a GPU on CE without specifying a custom image meant the workflow would never run because there wouldn't be a node available with GPU + CPU image.

# This PR's solution

Automatically adds the `-cuda` suffix to the default image when requesting a GPU.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 🐛 *Bug Fixes*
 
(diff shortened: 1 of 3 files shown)


SebastianMorawiec (approved):
💯

PR #372, opened 2024-03-12
imp: update sdk to pydantic v2

BenjaminMummery:
# The problem

Pydantic has been superseded by Pydantic V2. V2 provides significant advantages in efficiency.

# This PR's solution

1. Bumps the Pydantic requirement to `>=2.5`.
    - Ray 2.9.3 is compatible with Pydantic >=2.5, hence the lower bound.
2. Migrates code that uses pydantic to V2
    - Drop-in replacements:
        - `@validator`  -> `@field_validator`
        - `pydantic.error_wrappers.ValidationError`  -> `pydantic.ValidationError`
        - `pydantic.Field(*, regex="foo")` -> `pydantic.Field(*, pattern="foo")`
        - `GenericModel` -> `BaseModel`
        - `X: type = pydantic.Field(*, const = True)` -> `X: Literal[type] = pydantic.Field(*)`
        - `MODEL.dict()` -> `MODEL.model_dump()`[^1]
        - `MODEL.copy()` -> `MODEL.model_copy()`
        - `MODEL.json()` -> `MODEL.model_dump_json()`
        - `MODEL.parse_obj()` -> `MODEL.model_validate()`
        - `MODEL.parse_raw()` -> `MODEL.model_validate_json()`
        - `MODEL.schema_json()` -> `MODEL.model_json_schema()`
    - More complex changes
        - Optional fields without default values now cause validation errors when they're not provided. Solution: add `= None` default values for all optional fields.
        -  `MODEL.parse_file()` no longer exists, the file data has to be loaded then passed to `model_validate` or `model_validate_json`
        - `@validator(*, always=True)` no longer exists. By default, validators do not run on default values. This causes problems for sdk version checking where we want the validator to raise an exception if the version is None. The workaround is to change the field's type from `Optional[WorkflowMetadata]` to `Annotated[Optional[WorkflowMetadata], pydantic.Field(validate_default=True)]`.
3. Subclasses `BaseModel`
    -  There's a change in behaviour between Pydantic V1 GenericModel and Pydantic V2 BaseModel. This impacts our ability to load in Workflow results from previous SDK versions as the stored results cannot be depickled correctly. Specifically, `__pydantic_fields_set__` is missing from `BaseModel.__setstate__`.
    - We've solved this by subclassing `BaseModel` as `OrquestraBaseModel` and wrapping `__setstate__` to handle the missing key. 
    - If we stop supporting workflows from versions before whatever release this change is in, we should remove the subclass and go back to `BaseModel` (ticket for doing so: [ORQSDK-1025](https://zapatacomputing.atlassian.net/browse/ORQSDK-1025))
  
[^1]: This isn't a perfect replacement - `dict()` was shallow, `model_dump()` dumps nested models as well. Luckily this doesn't seem to affect our behaviour.

# Checklist

[ORQSDK-956]

[ORQSDK-956]: https://zapatacomputing.atlassian.net/browse/ORQSDK-956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ORQSDK-1025]: https://zapatacomputing.atlassian.net/browse/ORQSDK-1025?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pyproject.toml
```diff
@@ -27,7 +27,7 @@ requires-python = ">= 3.8"
 # the same API.
 dependencies = [
     # Schema definition
-    "pydantic>=1.10.8,<2",
+    "pydantic>=2.5, <3",
```
alexjuda, line 30: Is there any reason to restrict the top requirement? More info why we we should be open by default: https://bernat.tech/posts/version-numbers/

```suggestion
    "pydantic>=2.5",
```

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 💅 *Improvements*
 
 * Tracebacks in `orq` are made more compact to help with copy and pasting when an issue happens.
+* Bumped Pydantic version to `>=2.5.0,<3`
```
alexjuda, line 18:
```suggestion
* Bumped Pydantic version to `>=2.5.0`
```

src/orquestra/sdk/_base/_config.py
```diff
@@ -139,14 +139,16 @@ def _open_config_file() -> RuntimeConfigurationFile:
         raise exceptions.ConfigFileNotFoundError(
             f"Config file {config_file} not found."
         )
-    return RuntimeConfigurationFile.parse_file(config_file)
+    with open(config_file) as f:
+        data: str = f.read()
```
alexjuda, line 143:
```suggestion
    data: str = config_file.read_text()
```

src/orquestra/sdk/_base/_storage/_basemodel.py
```diff
@@ -0,0 +1,29 @@
...
+    """The pydantic BaseModel changed between V1 and V2.
+
+    As a result, workflow outputs generated prior to the V2 upgrade may not be
+    depickled correctly. The culpret is a change in behaviour of `__setstate__`.
+
+    This class adds a new `__setstate__` that wraps the V2 BaseModel `__setstate__` and
+    adds the missing behaviour back in.
+
+
+    """
```
alexjuda, line 20: **praise**: wow, what a great hack! The issue with depickling old objects sounded really bad; I'm glad you found a solution!

tests/runtime/ray/test_integration.py
```diff
@@ -911,7 +911,9 @@ def test_constants_and_inline_imports(runtime: _dag.RayRuntime):
         path_to_json = Path(__file__).parent.joinpath(
             "data/python_package/python_package_dependent_workflow.json"
         )
-        wf = ir.WorkflowDef.parse_file(path_to_json)
+        with open(path_to_json) as f:
+            data: str = f.read()
```
alexjuda, line 915:
```suggestion
        data = path_to_json.read_text()
```

alexjuda (approved):
🔥 

> Optional fields without default values now cause validation errors when they're not provided. Solution: add = None default values for all optional fields.

IIRC the previous behavior also caused type errors from pyright (which makes sense). It's a change for the better!

> `@validator(*, always=True)` no longer exists. 

IIRC this caused me a longer headache to get through. You found a much nicer workaround than I did.

> This isn't a perfect replacement - dict() was shallow, model_dump() dumps nested models as well. Luckily this doesn't seem to affect our behaviour. [↩](https://github.com/zapatacomputing/orquestra-workflow-sdk/pull/372#user-content-fnref-1-1a0fe42fee982a882d4d2b02dd2deb72)

Also seems like a change for the better!

src/orquestra/sdk/_base/_storage/_basemodel.py
```diff
@@ -0,0 +1,29 @@
+################################################################################
+# © Copyright 2024 Zapata Computing Inc.
+################################################################################
+
+from typing import Any
+
+from pydantic.main import BaseModel
+
+
+class OrquestraBaseModel(BaseModel):
```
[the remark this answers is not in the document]
BenjaminMummery: That's a really good question. I believe it would impact anything that we both pickle and write to a file (Results, WorkflowDefs, etc.) and anything that might be nested within them (TaskDefs for example). I started reasoning through this, but decided to go with a blanket change because:
1. find-replace is fast and easy
2. I explicitly intend this hack to be as short lived as possible, c.f. [ORQSDK-2025]
3. If we have a mix of pickle-safe OrquestraBaseModel and unsafe BaseModel, it becomes easier to accidentally add something unpicklable later. The error message this causes is _not_ particularly clear, and we run the risk of wasting time solving these issues.

src/orquestra/sdk/_base/_storage/_basemodel.py
BenjaminMummery: I can't claim all the credit, the internet helped a lot!

PR #389, opened 2024-04-16
Int: Cleanup remaining pyright exceptions

SebastianMorawiec:
# The problem
We use pyright but we still had bunch of files in exception list. 

# This PR's solution
Remove those files from exception list that involve our code.

# Checklist

### DIFF
--- a/src/orquestra/sdk/_client/_base/_ast.py
+++ b/src/orquestra/sdk/_client/_base/_ast.py
@@ -66,44 +66,49 @@ class _ReturnExprVisitor(ast.NodeVisitor):
         message: str,
     ):
         LOG.debug(message)
+        self._add_single()
+
+    def _add_single(self):
         self.outputs.add(_AstReturnMetadata(is_subscriptable=False, n_outputs=1))
 
-    def visit_Constant(self, node):
-        self.outputs.add(_AstReturnMetadata(is_subscriptable=False, n_outputs=1))
-
-    def visit_Name(self, node):
-        self.outputs.add(_AstReturnMetadata(is_subscriptable=False, n_outputs=1))
-
-    def visit_Attribute(self, node):
-        self.visit_Name(t.cast(ast.Name, node))
-
-    def visit_JoinedStr(self, node):
-        # f-strings
-        self.visit_Constant(t.cast(ast.Constant, node))
-
-    def visit_Set(self, node):
-        # Sets are python objects and are not subscriptable
-        self.outputs.add(_AstReturnMetadata(is_subscriptable=False, n_outputs=1))
-
-    def visit_Tuple(self, node):
-        self.visit_List(t.cast(ast.List, node))
-
-    def visit_List(self, node):
+    def _add_sequence(self, node, elements: t.List[ast.expr]):
         # If there is a starred node in the list, then unpacking is used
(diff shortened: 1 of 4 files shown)


src/orquestra/sdk/_client/_base/_ast.py
jamesclark-Zapata, line 1: I can't comment on the whole class to add a suggestion, but imo `typing.cast` is a last resort.

Wdyt about:

```
class _ReturnExprVisitor(ast.NodeVisitor):
    """The "inner" visitor.

    Handles any expression that's nested under the return statement.
    """

    def __init__(self):
        self.outputs: t.Set[_AstReturnMetadata] = set()

    def _fail_single(
        self,
        message: str,
    ):
        LOG.debug(message)
        self._add_single()

    def _add_single(self):
        self.outputs.add(_AstReturnMetadata(is_subscriptable=False, n_outputs=1))

    def _add_sequence(self, elements: t.List[ast.expr]):
        # If there is a starred node in the list, then unpacking is used
        if any(isinstance(e, ast.Starred) for e in elements):
            self._fail_single("Cannot infer number of outputs with unpacking")
            return

        n_elements = len(elements)
        if n_elements == 0:
            # Empty list counts as a single output, similarly to void functions
            # having a single `None` output.
            self._add_single()
        else:
            self.outputs.add(
                _AstReturnMetadata(is_subscriptable=True, n_outputs=n_elements)
            )

    def visit_Constant(self, node):
        self._add_single()

    def visit_Name(self, node):
        self._add_single()

    def visit_Attribute(self, node):
        self._add_single()

    def visit_JoinedStr(self, node):
        # f-strings
        self._add_single()

    def visit_Set(self, node):
        # Sets are python objects and are not subscriptable
        self._add_single()

    def visit_Tuple(self, node):
        self._add_sequence(node.elts)

    def visit_List(self, node):
        self._add_sequence(node.elts)

    def visit_Dict(self, node):
        # "None" inside the keys means there is dictionary unpacking being used
        # If None is used as a key, it is represented by an AST node
        if None in node.keys:
            self._fail_single("Cannot infer number of outputs with unpacking")
            return

        self._add_single()

    def generic_visit(self, node):
        self._fail_single(f"Assuming a single output for node {repr(type(node))}")
```

src/orquestra/sdk/_client/_base/cli/_entry.py
```diff
@@ -426,8 +426,9 @@ def status():
 )
 
 
+# Ignoring pyright errors as this class inherits typing issues form 3rd party code
 # region: login
-class GroupWithDefaultCommand(cloup.Group, DefaultGroup):
+class GroupWithDefaultCommand(cloup.Group, DefaultGroup):  # pyright: ignore
```
jamesclark-Zapata, line 431: **suggestion:** ‏Can we ignore specific errors? not sure if it works
```suggestion
# Ignoring pyright errors as this class inherits typing issues form 3rd party code
# region: login
class GroupWithDefaultCommand(
    cloup.Group, DefaultGroup
):  # pyright: ignore[reportIncompatibleMethodOverride,reportIncompatibleVariableOverride]
```

jamesclark-Zapata (approved):

PR #418, opened 2024-05-29
Int: Cleanup tests and dependencies of shared and runtime library

SebastianMorawiec:
# The problem
All the tests were left in SDK library.

# This PR's solution
Move proper tests to Shared and runtime library so it can be properly tested

# Checklist


projects/orquestra-sdk/tests/runtime/performance/test_cli_perf.py
alexjuda, line 1: Is there a reason why this file wasn't moved under `projects/orquestra-workflow-runtime/tests`?

projects/orquestra-sdk/tests/runtime/ray/conftest.py
alexjuda, line 1: Is there a reason why `projects/orquestra-sdk/tests/runtime/ray` directory wasn't moved under `projects/orquestra-workflow-runtime/tests`?

alexjuda (approved):
Found few potentially missed files

projects/orquestra-sdk/tests/runtime/performance/test_cli_perf.py
SebastianMorawiec: Yes - this test is testing CLI which is only part of SDK client side, its not available on runtime

projects/orquestra-sdk/tests/runtime/ray/conftest.py
SebastianMorawiec: As for anything left in runtime/ray - those tests are higher-level requiring us to use client-level concepts, like imports.  After we invent a way to generate IRs on flight inside RUNTIME package, we can move them all then, as now handling and maintanance would be very-very costly.

PR #430, opened 2024-06-28
docs: add section about backtracking errors

alexjuda:
# The problem

Installing dependencies can take a long time, and then fail, leaving the user clueless and frustrated. Unfortunately, there's not much we can do, because Ray seems to fail to capture ``pip`` output during the backtracking.

# This PR's solution

Adds an entry in our documentation about the problem, and a suggested fix scenario.

# Checklist


SebastianMorawiec (approved):
nice addition we can link to the users :)

PR #432, opened 2024-07-01
Fix: Fix that enables using orquestra-sdk without installing Runtime (and ray) at all

SebastianMorawiec:
# The problem
Using default orquestra-sdk installation was broken, as we still needed Ray to be installed as we had broken dependency graph
https://zapatacomputing.atlassian.net/browse/ORQSDK-1061

# This PR's solution
It introduces bunch of changes:

1. Most important one: Orquestra runtime is no longer required for usage of SDK (when not using Ray Runtime)
2. Make separation between SDK unit-tests and tests that require Ray Runtime - moved them all as Integration tests with separate github action
3. All SDK unit-tests are now being executed WITHOUT ray and orquestra-runtime being installed at all. This ensures that we never break 1st point again

# Checklist


.github/workflows/orquestra-sdk-integration.yml
```diff
@@ -0,0 +1,53 @@
+# This workflow runs integration tests
+
+# We need a workflow name to be able to schedule it from Github UI
+name: integration
```
alexjuda, line 4:
```suggestion
name: Integration Tests
```

.github/workflows/orquestra-sdk-integration.yml
```diff
@@ -0,0 +1,53 @@
+# This workflow runs integration tests
```
alexjuda, line 1: **nitpick**: could you paste here some of the reasoning you described in the PR description? The meaning of an _integration test_ is highly context-dependent.

For example, is the intention behind the tests here testing the integration between our packages (I guess this is the use case here), or the integration between components used _inside_ a single package (we have a `test_integration.py` file that starts Ray runtime and executes real workflows).

projects/orquestra-sdk/tests/integration/api/test_api_with_ray.py
```diff
@@ -5,7 +5,7 @@
 from unittest.mock import Mock
 
 import pytest
-from orquestra.workflow_runtime._testing import _connections
+from orquestra.workflow_runtime._testing import _connections  # type: ignore
```
alexjuda, line 8: What's the reason for this typing ignore? Is it importing a private symbol?

WDYT about actually making `orquestra.workflow_runtime.testing` public? To me this seems like a similar situation we had with _shared_. We made it public because this guards us against depending on private modules, which makes it easier to figure out when we introduce backwards-incompatible changes.

projects/orquestra-sdk/tests/integration/performance/test_cli_perf.py
```diff
@@ -20,7 +20,9 @@
 from pathlib import Path
 
 import pytest
-from orquestra.workflow_runtime._testing._connections import ray_suitable_temp_dir
+from orquestra.workflow_runtime._testing._connections import (  # type: ignore
```
alexjuda, line 23: Ditto private modules

projects/orquestra-sdk/tests/integration/ray/conftest.py
```diff
@@ -4,7 +4,7 @@
 import os
 
 import pytest
-from orquestra.workflow_runtime._testing import _connections
+from orquestra.workflow_runtime._testing import _connections  # type: ignore
```
alexjuda, line 7: Ditto private modules

projects/orquestra-sdk/tests/integration/ray/regression/test_workflow_outputs.py
```diff
@@ -5,8 +5,8 @@
 from typing import Any, Dict, Tuple, Union
 
 import pytest
-from orquestra.workflow_runtime._ray import _client, _dag
-from orquestra.workflow_runtime._testing import _connections
+from orquestra.workflow_runtime._ray import _client, _dag  # type: ignore
+from orquestra.workflow_runtime._testing import _connections  # type: ignore
```
alexjuda, line 9: Ditto private modules

alexjuda (commented):
Makes sense to me overall. Added one nitpick and a question about private modules I'd want to discuss before moving on

projects/orquestra-sdk/tests/integration/api/test_api_with_ray.py
SebastianMorawiec: this # type: ignore is not because we have any issues with private/public modules.
Its comming from the fact that style-check we use do not have orquestra.workflow_runtime installed right now, so it marks all usages of it as invalid.

This can be fixed in 2 ways:
Easy way - is what I did
Hard way - make style-fix install orquestra.workflow_runtime - which is hard as we rely on the fact that it uses the same makefile recipes as test-coverage, which we intentionally do NOT want to have orquestra.workflow_runtime installed

SebastianMorawiec: Ill create a ticket to transition to the "hard way". Now when im thinking about it, its better and more reliable.

SebastianMorawiec: Nvm, agreeed offline how to proceed with this PR

alexjuda (approved):
Added non-blocking suggestions

PR #438, opened 2024-07-23
Fix: revert dependency to ray[default]

SebastianMorawiec:
# The problem
When running ray workflows without installing [default] extra, ray behaves weirdly - some workflow fail without any particular reason, other are stuck.
This all was tracked down to change we did to reduce number of dependnecies

# This PR's solution
Revert the change - always install ray[default] in our system as we use ray only to run workflows.

# Checklist

### DIFF
--- a/projects/orquestra-workflow-runtime/pyproject.toml
+++ b/projects/orquestra-workflow-runtime/pyproject.toml
@@ -50,6 +50,7 @@ ray = [
     "async-timeout",       # Ray depends on this, but its missing from ray requirements. Can be removed when https://github.com/ray-project/ray/issues/41267 is fixed.
 ]
 
+# this extra is left for backward-compatibility reasons.
 dashboard = [
     "orquestra-workflow-runtime[ray]",
 ]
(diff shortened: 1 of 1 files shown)


projects/orquestra-workflow-runtime/pyproject.toml
```diff
@@ -45,14 +45,13 @@ dependencies = [
 
 [project.optional-dependencies]
 ray = [
-    "ray==2.30.0",
+    "ray[default]==2.30.0",
     "virtualenv>=20.0.24, !=20.21.1",  # Ray depends on this, but it's missing from the base requirements.
     "async-timeout",       # Ray depends on this, but its missing from ray requirements. Can be removed when https://github.com/ray-project/ray/issues/41267 is fixed.
 ]
 
 dashboard = [
```
jamesclark-Zapata, line 53: **nitpick:** ‏we no longer need this optional extra. Should we leave it for backwards compat or remove it?

jamesclark-Zapata (approved):
Failing checks look unrelated.

LGTM

projects/orquestra-workflow-runtime/pyproject.toml
SebastianMorawiec: Yea we use it in SDK package as well, so for backward compat I left it. Ill add a comment as well why it is here

PR #445, opened 2024-09-03
Int: Fix newest flake8 problems in SDK

SebastianMorawiec:
# The problem
Flake8 started to detect problems with our docstrings

# This PR's solution
I generally tried to fix some one the issues, but this error is kind of weird - it prevents you from adding docstrings for exceptions that are raised by sub-functions or inehritance which is misleading.
I fix them for 2 of our libraries and ignored for SDK mostly, since fixing this problem would make our code worse than it was IMHO

# Checklist

### DIFF
--- a/projects/orquestra-sdk/pyproject.toml
+++ b/projects/orquestra-sdk/pyproject.toml
@@ -250,6 +250,7 @@ ignore = [
     'DOC201', # Function/method does not have a return section in docstring
     'DOC203', # Return type(s) in docstring not consistent with the return annotation
     'DOC301', # `__init__()` should not have a docstring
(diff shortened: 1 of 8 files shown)


alexjuda (commented):
Please run a build of docs and check if the API reference looks okay. I remember adding fully-qualified module + class names for exceptions because otherwise sphinx would resolve invalid class names in the docs

SebastianMorawiec:
> Please run a build of docs and check if the API reference looks okay. I remember adding fully-qualified module + class names for exceptions because otherwise sphinx would resolve invalid class names in the docs

Ive checked... and it looks like API reference is already broken on the latest version of SDK anyway. We have to fix that independently

alexjuda (approved):

